PR #8, opened 2020-12-07
[CORE-1309] Smart App Shell

msanroman:
## Purpose

To implement the global organizations switcher following the smart App Shell approach we've discussed in the frontend syncs.

More details [here](https://paper.dropbox.com/doc/Integrating-with-the-Smart-App-Shell--BA2L7UecyaX4Lw8MN~RB48ohAg-SBPXfQhWSJwgBb3yhMEQJ).

## Review details

You can pretty much ignore the files under `packages/app` - this is a create-react-app for testing purposes only. It'd be great to follow up and actually use a frontend similar to how we build our own with a Webpack dev server and very little else.


packages/app-shell/src/graphql/account.js
```diff
@@ -0,0 +1,33 @@
...
+  mutation AccountSetCurrentOrganization($organizationId: String) {
+    accountSetCurrentOrganization(organizationId: $organizationId)
+  }
+`
+
+export const QUERY_ACCOUNT = gql`
+  query GetAccount {
+    account {
+      id
+      email
```
anafilipadealmeida, line 13: @msanroman, could we also query the Customer name (https://github.com/bufferapp/buffer-api-gateway/blob/main/src/schema/schema-core-api.ts#L85) to use in the NavBarMenu? I believe user.name is currently not getting anything
Screenshot 2021-01-05 at 17 32 13

hamstu: :ceilingcat: I don't think we actually store the customer name in the account DB. 🤔

msanroman: Yeah, the `Customer` type is for Billing (so an Organization has one customer, e.g: one Stripe account) and the `Account` type has no field for the name 🤔 

I believe all products (except Publish) use the App Shell without the name. Not sure if that's the pattern we want to follow?

anafilipadealmeida: Thank you both! Looping @letsbleachthis on this one, and I'll also bring it to the Publish sync 🙂

packages/app-shell/src/NavBar/index.jsx
```diff
@@ -204,145 +238,150 @@ export function appendOrgSwitcher(orgSwitcher) {
...
+          menubarItem={
+            <NavBarMenu user={user} isImpersonation={user.isImpersonation} />
+          }
+          items={[
+            ...appendOrgSwitcher(user, switchOrganization, channels),
+            appendMenuItem(ignoreMenuItems, {
+              id: 'account',
+              title: 'Account',
+              icon: <PersonIcon color={gray} />,
+              hasDivider: false,
```
anafilipadealmeida, line 341: **Current:**
Screenshot 2021-01-07 at 13 15 03
**Before:**
Screenshot 2021-01-07 at 13 16 05
@msanroman, I just noticed the hasDivider in the account button has become false, was it intended? It was `orgSwitcherHasItems` before, and would create a clear division between the list of orgs and non-related buttons

packages/app-shell/src/graphql/account.js
letsbleachthis: Thanks for the ping, @anafilipadealmeida! Was just checking the experience in each of the products, and it looks like I see my name in Analyze too 🤔  https://share.buffer.com/z8u49OnK

Reaching back into the depths of my brain to recall our early conversations about users & permissions for One Buffer (we started that work, then put it on pause to prioritize shared channels instead)...

We had decided that we do want to capture and store a user's name at _some_ point (either in the invite flow or perhaps onboarding). There are a few reasons I think it makes sense, the best of which is probably to improve the collaboration experience (tagging Kamala Harris in a comment is more readable than kamala.harris@thewhitehouse.com). Sharing in case that future state is helpful, but for the time being, if Publish _is_ the outlier, then I think it would be okay to remove the name for consistency's sake.

packages/app-shell/src/NavBar/index.jsx
```diff
@@ -204,145 +232,113 @@ export function appendOrgSwitcher(orgSwitcher) {
...
+    helpMenuItems,
+    onLogout,
+    displaySkipLink,
+    onOrganizationSelected,
+    menuItems,
+    ignoreMenuItems,
+    channels,
+  } = props;
+
+  const user = useContext(UserContext);
```
hamstu, line 247: One idea to make this context a bit easier to consume would be to make and export our own custom hook from the `app-shell` that looks something like this:

```js
function useUser() {
  const context = React.useContext(UserContext)
  if (context === undefined) {
    throw new Error('useUser must be used within a UserContext.Provider')
  }
  return context
}
```

Inspired by [this article from KCD](https://kentcdodds.com/blog/how-to-use-react-context-effectively).

hamstu (approved):
This is great @msanroman! Thanks so much for all the hard work you've put in here.

packages/app-shell/src/NavBar/index.jsx
msanroman: Great point! I forgot to take that extra step!

Thanks @hamstu ! 🌮

PR #10, opened 2021-01-21
Update 'Engagement' URL

ssvmvss:
## Overview

This PR has the changes needed to always make the `Engagement` link point to `https://engage.buffer.com`, removing the decision of redirecting the user to the signup page if needed. 

Initially users without an `engage` product link would be redirected so they can perform a signup first but, we have an extra step now, based on the `canAccessEngagement` value so users under a free plan are redirected to `https://engage.buffer.com/free-plan`. Ideally this kind of decisions should be made on the App Shell to avoid extra redirects but that's not easy for now. We will make that improvement once the smart App Shell is ready 🙂

### DIFF
--- a/src/NavBar/NavBarProducts/NavBarProducts.jsx
+++ b/src/NavBar/NavBarProducts/NavBarProducts.jsx
@@ -142,10 +142,10 @@ const getLogo = (product) => {
 
 const NavBarProducts = ({ products, activeProduct }) => (
   <StyledNavBarProducts>
-    {products.map(({ id, label, href, isNew }) => (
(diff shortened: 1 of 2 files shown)


gmzjuliana (approved):
LGTM thanks @ssvmvss

ssvmvss:
Heya @anafilipadealmeida, @gmzjuliana and @albennett! 👋 Today we've realized we need this change on the AppShell in order to launch Engage on Monday / Tuesday successfully. I'll go ahead with this merge without waiting for the reviews to happen so I can move forward with getting everything ready for Monday morning. Tagging you here anyway so you are aware of this! 🙂

PR #13, opened 2021-02-23
[PUB-3200] Modify org switcher setCurrentOrganization mutation

anafilipadealmeida:
Modify setCurrentOrganization mutation, to update the cache
https://buffer.atlassian.net/browse/PUB-3200

Extra:
- Remove apolloclient from dependencies (it would cause a mismatch version issue, just like react)


packages/app-shell/src/graphql/account.js
```diff
@@ -28,9 +28,11 @@ export const QUERY_ACCOUNT = gql`
       organizations {
         id
         name
-        canEdit
```
anafilipadealmeida: @albennett , these 3 params need to stay for now, I'll add them back 😅

albennett (approved):
This looks great @anafilipadealmeida . Maybe we can add a front-end task to revisit the test issues.

packages/app-shell/test/NavBar.test.jsx
```diff
@@ -1,122 +1,141 @@
...
-                  id: account.organizations[1].id,
-                }
-              }
-            }
-          }]}>
-            <UserContext.Provider value={account}><NavBar onOrganizationSelected={onOrganizationSelected}/></UserContext.Provider>
-          </MockedProvider>,
-        );
-      }
+    // describe ('changing organizations', () => {
```
gmzjuliana, line 73: hey @anafilipadealmeida i've just now seen these comments, do we need them still or how come we kept them but commented?

PR #15, opened 2021-02-23
change host url to 0.0.0.0

federicoweber:
This is needed so that the dev server can be piped from a VM, the app shell will still be available on `appshell.local.buffer.com`


msanroman:
This always gets me 😂 Thanks, @federicoweber !

PR #17, opened 2021-03-02
Task/plan selector first screen

gmzjuliana:
Adds the plan selector to the modals
Conditionally renders appropriate header, cards, button with next action template
Conditionally renders summary with appropriate bullet points and price


packages/app-shell/src/graphql/account.js
```diff
@@ -19,10 +19,59 @@ export const QUERY_ACCOUNT = gql`
...
+        }
       }
       organizations {
         id
         name
+        billing {
+          canAccessAnalytics
+          canAccessEngagement
+          canAccessPublishing
+        }
```
federicoweber, line 74: Why to we need this informations at this level?

federicoweber (commented):
Hey @gmzjuliana this is looking great, I left a feel small comments and noticed a small visual bug (see attached), but other than that it's looking solid.

Let's wire it up to the `updatePaymentMethod` and the `selectPaymentMethod` modal and we are ready to ship it.

Great work 🌮.

image

PR #18, opened 2021-03-02
Payment Method Modal

federicoweber:
To add the Payment Method Modal
image
[(ref.)](https://www.figma.com/file/D81vy7rYgF1dd8xvV8UsrQ/Buffer-Plan-Selector?node-id=651%3A37993)

### DIFF
--- a/packages/app-shell/package.json
+++ b/packages/app-shell/package.json
@@ -29,7 +29,7 @@
     "@babel/plugin-transform-runtime": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
-    "@bufferapp/ui": "^5.48.2",
+    "@bufferapp/ui": "^5.6",
     "@storybook/addon-actions": "^6.1.19",
     "@storybook/addon-essentials": "^6.1.19",
     "@storybook/addon-links": "^6.1.19",
@@ -54,7 +54,7 @@
   },
   "peerDependencies": {
     "@apollo/client": "^3.3.4",
-    "@bufferapp/ui": ">= 5.48.2",
+    "@bufferapp/ui": ">= 5.6",
     "prop-types": ">= 15.7.2",
     "react": ">= 16.8.0",
     "react-dom": ">= 16.8.0",
--- a/packages/app-shell/package.json
+++ b/packages/app-shell/package.json
@@ -29,7 +29,7 @@
     "@babel/plugin-transform-runtime": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
-    "@bufferapp/ui": "^5.60",
+    "@bufferapp/ui": "^5.60.1",
     "@storybook/addon-actions": "^6.1.19",
     "@storybook/addon-essentials": "^6.1.19",
     "@storybook/addon-links": "^6.1.19",
--- a/packages/app-shell/package.json
+++ b/packages/app-shell/package.json
@@ -29,7 +29,7 @@
     "@babel/plugin-transform-runtime": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
-    "@bufferapp/ui": "^5.6",
+    "@bufferapp/ui": "^5.60",
     "@storybook/addon-actions": "^6.1.19",
     "@storybook/addon-essentials": "^6.1.19",
     "@storybook/addon-links": "^6.1.19",
@@ -54,7 +54,7 @@
   },
   "peerDependencies": {
     "@apollo/client": "^3.3.4",
-    "@bufferapp/ui": ">= 5.6",
+    "@bufferapp/ui": ">= 5.60",
     "prop-types": ">= 15.7.2",
     "react": ">= 16.8.0",
     "react-dom": ">= 16.8.0",
--- a/packages/app-shell/package.json
+++ b/packages/app-shell/package.json
@@ -62,6 +62,8 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.4",
+    "@stripe/react-stripe-js": "^1.3.0",
+    "@stripe/stripe-js": "^1.13.0",
     "graphql-tag": "^2.11.0"
   },
   "gitHead": "29cc1e0b448417d26aa0bf47eb4eda466641e4bf"
--- a/packages/app-shell/package.json
+++ b/packages/app-shell/package.json
@@ -62,6 +62,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.4",
+    "@apollo/react-testing": "^4.0.0",
     "@stripe/react-stripe-js": "^1.3.0",
     "@stripe/stripe-js": "^1.13.0",
     "graphql-tag": "^2.11.0"
--- a/packages/app-shell/src/Modal/index.jsx
+++ b/packages/app-shell/src/Modal/index.jsx
@@ -9,6 +9,8 @@ const ModalContent = ({ modal }) => {
       return (<PaymentMethod />);
     case MODALS.planSelector:
       return (<div> Plan Selector </div>);
+    case MODALS.success:
+      return (<div> Success </div>);
     default:
       return null;
   }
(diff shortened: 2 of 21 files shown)


packages/app-shell/src/Modal/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+    case MODALS.paymentMethod:
+      return (<PaymentMethod />);
+    case MODALS.planSelector:
+      return (<div> Plan Selector </div>);
+    default:
+      return null;
+  }
+}
+
+const Modal = ({ modal, openModal }) => {
```
federicoweber, line 17: Hey @gmzjuliana this is the Component that's responsible to render the various modals.

packages/app-shell/src/PaymentMethod/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+  >
+    <Content openPlans={openPlans} />
+  </Elements>)
+}
+
+const PaymentMethod = () => {
+  return (
+    <UserContext.Consumer>{user => (
+      <ModalContext.Consumer>{modal => (
+        <StripeProvider user={user} openPlans={() => {modal.openModal(MODALS.planSelector)}}/>
```
federicoweber, line 121: This is how we can render one modal from another one, will extend this so that we can pass a `data` object that we can use to pass data, and components across the modals.

It will be something like this `modal.openModal(MODALS.planSelector, data)`
The idea there is that we can use this to pass things like: the selected plan, and the entire summary component from the `PlanSelector`, to the `PaymentMethod`; and from the `PaymentMenthod` to the `Confirmation`.

packages/app-shell/src/hooks/useModal.test.js
```diff
@@ -54,5 +54,15 @@ describe('useModal', () => {
 
       expect(result.current.modal).toEqual(MODALS.planSelector)
     })
+
+    it('set a modal with data', () => {
+      const { result } = renderHook(() => useModal())
+      const data = { foo: 'bar' }
+      act(() => {
+        result.current.openModal(MODALS.planSelector, data)
```
federicoweber, line 62: This is how we can pass around data between modals 🙂 .

gmzjuliana (approved):
hey @federicoweber this looks great, I left some question to make sure I understand the way that data gets passed around, I'd recommend reading them all before you reply cause I think I answer some of my own questions 😂  I also left a comment on the accessibility of the error message 😊

packages/app-shell/src/PaymentMethod/index.jsx
federicoweber: It will be in the context

packages/app-shell/package.json
```diff
@@ -29,7 +29,7 @@
     "@babel/plugin-transform-runtime": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
-    "@bufferapp/ui": "^5.48.2",
+    "@bufferapp/ui": "^5.6",
```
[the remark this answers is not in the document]
federicoweber: Not sure I got this 😕. What do you mean?

packages/app-shell/src/PaymentMethod/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+  useEffect(() => {
+    console.log(stripeResponse);
+  }, [stripeResponse])
+
+  return (<StyledForm>
+    <LeftSide>
+      <Text type='h2'>Billing Details</Text>
+      {stripeResponse && stripeResponse.error && stripeResponse.error.type === "invalid_request_error" &&
+        <><Error error={stripeResponse.error} /><br/></>
```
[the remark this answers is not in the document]
federicoweber: ah good one, let's chat about that

packages/app-shell/src/hooks/useModal.js
```diff
@@ -17,14 +18,19 @@ function useModal() {
     }
   }, [])
 
-  const openModal = useCallback((modalKey) => {
+  const openModal = useCallback((modalKey, modalData = null) => {
     const matchingModal = Object.keys(MODALS).find(k => k === modalKey)
     if (matchingModal) {
       setModal(matchingModal)
+      if (modalData) {
+        setData(modalData)
```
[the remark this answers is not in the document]
federicoweber: yep that's right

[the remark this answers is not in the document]
federicoweber: and this commit actually have all the changes needed to get pass that around, so we are fully operational 😉

packages/app-shell/package.json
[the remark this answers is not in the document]
gmzjuliana: 😅 I assume here that since we're at version 5.60 that we can omit the 0 on the 60? but what happens here since we have a version 5.6 published months ago? during the install, how does the compiler know that you want 5.60 and not 5.6?

packages/app-shell/src/PaymentMethod/hooks/useUpdateSubscriptionPlan.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react'
+import { useMutation } from '@apollo/client';
+import { UPDATE_SUBSCRIPTION_PLAN } from '../../graphql/billing'
+
+function useUpdateSubscriptionPlan({ user, plan, interval, hasPaymentMethod }) {
```
federicoweber, line 5: hey @gmzjuliana I've added in here the hook to update the selected plan, if the user has a payment method

packages/app-shell/src/PaymentMethod/hooks/useUpdateSubscriptionPlan.test.jsx
```diff
@@ -0,0 +1,158 @@
+import React from 'react';
```
federicoweber, line 1: Also you can take a look at this as an example of testing hooks in the case where the component need to use a mutation or a query

packages/app-shell/src/PaymentMethod/hooks/useUpdateSubscriptionPlan.test.jsx
```diff
@@ -0,0 +1,158 @@
...
+      },
+      error: new Error('The horror! The horror!')
+    }
+  ]
+
+  function testHook(params) {
+    const { result, waitForNextUpdate } = renderHook(
+      () => useUpdateSubscriptionPlan(params),
+      {
+        wrapper: ({ children }) => (
```
federicoweber, line 61: This is were the magic happens, we are telling renderHook to wrap it's test component with an Apollo provider

packages/app-shell/src/PaymentMethod/hooks/useCreatePaymentMethod.js
```diff
@@ -0,0 +1,70 @@
+import { useEffect, useState } from 'react';
+import {
+  useStripe,
+  useElements,
+  CardNumberElement,
+} from "@stripe/react-stripe-js";
+
+function useCreatePaymentMethod(setupIntent) {
```
federicoweber, line 8: I will add some tests for this one before merging it in

packages/app-shell/package.json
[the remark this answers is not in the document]
federicoweber: oh great catch, `5.6` should be a different version then `5.60`, not sure what is going on here, I've fixed it in the package

packages/app-shell/package.json
```diff
@@ -29,7 +29,7 @@
     "@babel/plugin-transform-runtime": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
     "@babel/preset-react": "^7.12.1",
-    "@bufferapp/ui": "^5.48.2",
+    "@bufferapp/ui": "^5.60",
```
gmzjuliana, line 32: we can update this to 5.60.1 so the modal doesn't have that border

packages/app-shell/src/PaymentMethod/hooks/useUpdateSubscriptionPlan.js
gmzjuliana: nice! thanks fed

gmzjuliana (approved):
looks good! thanks Fed and thanks for leaving the hints around, will definitely come in handy in my implementation

PR #20, opened 2021-03-10
Add Confirmation Screen

gmzjuliana:
Adds the success modal for 
- team/individual plan change
- change to free
- starting a new trial
- changing billing info

also passes data for the success modal
Also adds a useInterval hook to render the plan selector with the correct initialization
fixes header spacing

### DIFF
--- a/packages/app-shell/src/Confirmation/hooks/useCopy.js
+++ b/packages/app-shell/src/Confirmation/hooks/useCopy.js
@@ -1,0 +1,36 @@
+const useCopy = ({ planName, startedTrial, onlyUpdatedCardDetails }) => {
+  if (startedTrial) {
+    return {
+      label: 'Congrats! You are now starting your trial',
+      description:
+        'For the next 14 days you get to experience Buffer to it’s full. Have fun!',
+      buttonCopy: "Great. Let's Go",
+    };
+  }
+
+  if (onlyUpdatedCardDetails) {
+    return {
+      label: 'Your billing details are now updated',
+      description:
+        'Thank you, your billing details have gone through successfully.',
+      buttonCopy: "That's great!",
+    };
+  }
+
+  if (planName === 'Free') {
+    return {
+      label: 'Congrats! You are now on the Free plan',
+      description:
+        'Your change have gone through successfully. Start using your Free plan today.',
+      buttonCopy: "Great. Let's Go",
+    };
+  } else {
+    return {
+      label: `Congrats! Welcome to the ${planName} plan`,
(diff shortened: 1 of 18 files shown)


packages/app-shell/src/Confirmation/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+const Confirmation = () => {
+  return (
+    <UserContext.Consumer>
+      {(user) => (
+        <ModalContext.Consumer>
+          {({ openModal, data }) => (
+            <Screen
+              selectedPlan={mockSelectedPlan} //data.selectedPlan
+              onlyUpdatedCardDetails={onlyUpdatedCardDetails} //data.onlyUpdatedCardDetails
+              startedTrial={startedTrial} //data.isTrial
```
gmzjuliana, line 85: hey @federicoweber i used mocks in here to be able to test it but on Monday we can make the changes in the paymentMethod/startTrialModal to hook it up so we pass the selectedPlan and 2 bools for onlyUpdatedCardDetails and startedTrial

packages/app-shell/src/Confirmation/hooks/useCopy.js
```diff
@@ -0,0 +1,36 @@
...
+  if (startedTrial) {
+    return {
+      label: 'Congrats! You are now starting your trial',
+      description:
+        'For the next 14 days you get to experience Buffer to it’s full. Have fun!',
+      buttonCopy: "Great. Let's Go",
+    };
+  }
+
+  if (onlyUpdatedCardDetails) {
```
federicoweber, line 11: Do you think we can remove the `onlyUpdatedCardDetails` check and just set it as the default option, in case there is no plan and `startedTrial` is false?

packages/app-shell/src/Confirmation/hooks/useCopy.js
```diff
@@ -0,0 +1,36 @@
+const useCopy = ({ planName, startedTrial, onlyUpdatedCardDetails }) => {
```
federicoweber, line 1: this feels more like a `getCopy` function then an hook to me, as we are not using any react specific logic in here (states, ref, …) 😉

packages/app-shell/src/Confirmation/index.jsx
federicoweber: I see you updated the callbacks but you are still using the test values in here 😉.

federicoweber (changes requested):
This is looking great @gmzjuliana, I left a few small comments, but the only blocker is with the test data we are passing to the confirmation modal :).

Feel free to ship this once that it's ok.

Thanks!

PR #21, opened 2021-03-11
[AT-866] Add Updgrade, Start trial CTA and Start trial modal

federicoweber:
To show a CTA in the Navigation bar to prompt free user to upgrade, or start a trial if they are elegible


packages/app-shell/src/StartTrial/style.js
```diff
@@ -0,0 +1,63 @@
...
+
+export const Holder = styled.div`
+  display: flex;
+  width: 800px;
+  height: 376px;
+  box-sizing: border-box;
+  background-repeat: no-repeat;
+  background-position-x: right;
+  background-position-y: bottom;
+  background-image: url(https://buffer-ui.s3.amazonaws.com/Confirmation+Screen+-+Background.png);
```
gmzjuliana, line 18: suuuuuuuper small, can we put the url in a string so it formats nice? 😬

PR #23, opened 2021-03-15
add a nicer loading state

gmzjuliana:
Adds a nicer loading as we wait for the fetch to finish. This is really boilerplate! Looking for a bit more design advice still https://share.buffer.com/E0u90Aj7

also fixes storybook for plan selector and summary


packages/app-shell/src/PlanSelector/components/Loader.jsx
```diff
@@ -0,0 +1,64 @@
...
+  Right,
+  Left,
+  Wrapper,
+  Container,
+  CardContainer,
+  LoadingSummary,
+  LoadingText,
+} from '../style';
+
+const Loader = () => {
```
federicoweber, line 18: Hey @gmzjuliana, this is looking fantastic but I wanted to share some reflection on the architecture of the loading state 🙂.

This is mimicking the structure of the plan selector modal quite well and in terms of  user experience it's working great but I can see two problems with this approach:
1. the loader is a separate component, so we are effectively re-rendering the entire modal once we get the data, and have can have a big performance cost, especially if we want to add in animations.
2. if we change the layout of the modal we will need to update the layout of the loader as well

As an alternative approach, I would recommend pushing the responsibility of handling the loading state to the individual components: `plan cards`, `summary, and `buttons`. We won't even need to pass down the loading prop, we can infer that from their expected props state, and have those elements show the loading state if they are missing the data.
This approach will guarantee that the layout will always match, as we will only have one, plus will greatly improve performances because we will re-render smaller components, without causing a big reflow.

Let me know if you want to chat more about this 🙂.

gmzjuliana:
this is super outdated now and it seems like we might try another approach. gonna close this for now so we keep things tidy 🧹

PR #27, opened 2021-03-22
[AT-871] Adds plan selector viewed segment tracking

dinostheo:
Tracks the `plan selector viewed` & `page viewed`, when the plan selector is opened by a non impersonated user.
https://buffer.atlassian.net/jira/software/projects/AT/boards/69?selectedIssue=AT-871

There are some CTA related properties that we will have to discuss how to make the available to the tracking. 
Also, the `window.PRODUCT_TRACKING_KEY` is exposed for all products except `Engage`, we might have to expose that in engage or find a different way to get the product from the url perhaps.


gmzjuliana (approved):
hey @dinostheo I know you still have the CTAs to fill in but the rest looks good to me! thanks for working on this 🌮

dinostheo:
Hey @gmzjuliana 
Would you be up taking another look in this PR? I added some CTAs. I found that we have a data parameter on the `openModal` method, which can then retrieve from the `ModalContext` and I'm using it to pass the CTAs of one page to another. 
It works, but it might not be the best way to do it. I would appreciate your feedback.

gmzjuliana (approved):
ooooh such a good idea to use the data from the modals 👏 great thinking @dinostheo

dinostheo:
Tested with segment
Screen Shot 2021-03-25 at 09 45 26

I will merge it.

PR #28, opened 2021-03-22
Redirect to logout url if account has an error state

albennett:
Handle the redirect to login in the app-shell if the user does not have an account

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+
+## [2.6.0] - 2021-03-24
+- Redirect account 400 error to login
+
 ## [2.5.1] - 2021-03-24
 - Add missing fields in account query
 
--- a/packages/app-shell/src/index.jsx
+++ b/packages/app-shell/src/index.jsx
@@ -3,14 +3,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { useQuery } from '@apollo/client';
 
-import NavBar from './NavBar';
+import NavBar, { getLogoutUrl } from './NavBar';
 import Banner from './Banner';
 import Modal from './Modal/index';
 
 import { AppShellStyled, ContentWrapper, SidebarWrapper, Wrapper } from './style';
 import { UserContext } from './context/User';
 import { ModalContext } from './context/Modal';
-import useOrgSwitcher from './hooks/useOrgSwitcher';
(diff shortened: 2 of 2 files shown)


gmzjuliana (approved):
This looks good to me @albennett, I'm curious though if someone from product has approved it? It might be a bit jarring to get completely kicked out of the app 🤔

albennett:
Hey @gmzjuliana great one, I haven't reached out to product since this is currently how our logic works. With Engage, it happens [here](https://github.com/bufferapp/buffer-engage-client/blob/79bc425a0a686791bf3180e50e2643439efce9cd/src/modules/account/hooks/useAccountCredentials/index.js#L59).

With Analyze and Publish, we've been using this unauthorized redirect package: https://github.com/bufferapp/buffer-js-unauthorized-redirect

Now that we're displaying missing access pages, this redirect package would cause a redirect if a user has an account but is missing access to a particular product. This is why we're removing this unauthorized redirect package in publish and instead trying to handle the redirect for a missing account only.

ssvmvss (approved):
Heya @albennett! 👋 This looks good to me! ❤️  The only difference I see with Engage is that we would redirect the user to the Login screen skipping the logout part but I think it makes sense! As far as I understand a request to the API Gateway would only return a `400` error if the user does not have the credentials that are needed so I think we should be OK!

I really like the idea of trying to get as much consistency as possible between solutions and now that the AppShell is smart and has access to all the account data it sounds like a great place to handle all this kind of things! 🙂

albennett:
Thanks @ssvmvss it does feel like a 401 error would make more sense here. I'll bring it up to Core in case a future change makes sense.

ssvmvss:
Sounds good @albennett! 🙌

ssvmvss:
I just checked on Engage @albennett to see what the error details were for `400` and what I can see is the following:

Screen Shot 2021-03-24 at 19 53 03

When it should be failing with an "unauthorized" error instead (usually `401` as you mentioned!).

Leaving this here in case you want to provide extra details to Core! 🙂

gmzjuliana:
hey @albennett I wanted to leave a comment about something I noticed working with this version. I'm working in my local to extend getAccount and if I mess up the query, it'll kick me out completely instead of showing me the error in the query. I mentioned it to @ssvmvss and she tipped me off that maybe core send 400 in general, not only for unauthorized 🤔  What do you think?

PR #32, opened 2021-03-25
Plan Selector QA 4

gmzjuliana:
see the commits for details :)

### DIFF
--- a/packages/app-shell/src/PaymentMethod/components/Form.jsx
+++ b/packages/app-shell/src/PaymentMethod/components/Form.jsx
@@ -11,10 +11,10 @@ import {
   LeftSide,
   RightSide,
   Error,
-  ButtonContainer
-} from '../style'
-import Field from './Field'
-import Summary from '../../Summary'
+  ButtonContainer,
+} from '../style';
+import Field from './Field';
+import Summary from '../../Summary';
 
 import useSetupIntent from '../hooks/useSetupIntent';
 import useCreatePaymentMethod from '../hooks/useCreatePaymentMethod';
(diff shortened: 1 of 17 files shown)


packages/app-shell/src/PaymentMethod/components/Form.jsx
```diff
@@ -90,7 +104,7 @@ const Form = ({
     <RightSide>
       {plan && <Summary selectedPlan={plan}/>}
       <ButtonContainer>
-        <Button type="primary" onClick={submit} disabled={!submitEnabled || processing} label={processing ? "Processing..." : plan ? "Confirm Payment" : 'Update Payment Details'} fullWidth />
+        <Button type="primary" onClick={submit} disabled={!submitEnabled || processing} label={getButtonLabel()} fullWidth />
```
dinostheo, line 107: 😍

packages/app-shell/src/Summary/index.jsx
```diff
@@ -30,42 +35,54 @@ const Summary = ({
...
     : '';
 
+  const isDowngrading = (currentPlanId, selectedPlanId) => {
+    if (currentPlanId === 'individual') {
+      return selectedPlanId === 'free' ? true : false;
+    }
+    if (currentPlanId === 'team') {
+      return true;
+    }
+    if (currentPlanId === 'free') {
```
dinostheo, line 45: The following comment is not blocking and won't change anything in the way the codebase works. I was just thinking that it might be a good idea to Create a plans module, where we set the Plan IDs and potentially other information there once and use them through its interface.
e.g. the above line and similar lines of code could become
```
currentPlanId === plan.freePlanId
```

Naming might not be great here, but I hope it transfers my thinking 😅

dinostheo: Potentially, methods related to plans e.g. isDowngrade, isUpgrade could be defined once in this module.

dinostheo (approved):
It looks good to me @gmzjuliana.
I left a minor comment, regarding plan and checking against specific plan values, but it is not blocking.

packages/app-shell/src/PaymentMethod/index.jsx
```diff
@@ -45,6 +45,10 @@ const PaymentMethod = () => {
                   }}
                   user={user}
                   plan={data ? data.plan : null}
+                  isTrial={
+                    user.currentOrganization.billing.subscription.trial
```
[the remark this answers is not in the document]
gmzjuliana: good catch, thanks for pointing it out !

packages/app-shell/src/Summary/index.jsx
gmzjuliana: this is a good idea, @dinostheo :) I've been thinking about how to clean up some parts of the code and hope to find some time to do so, this might do just that! I'm gonna leave it out if this PR just for times sake but will revisit asap, thanks

PR #36, opened 2021-03-30
[Fix] Check if previous data is defined when updating cache after org

dace:
### Description

There seems to be an error related to updating the cache with `previousData` when a user selects a new org in the org switcher ([Slack thread for context](https://buffer.slack.com/archives/C01K9FWJVBJ/p1617122335356700)):

e257993087e76d578c2237d5195595b1

@anafilipadealmeida @msanroman I am unfamiliar with updating the cache, so please let me know if this can be better handled (for example, if we should expect `previousData` to always be defined). 

### Screenshots

**Before:**

https://user-images.githubusercontent.com/10112294/113049861-23ed8500-9172-11eb-92c7-95321a52c9af.mp4

**After:**

https://user-images.githubusercontent.com/10112294/113049873-27810c00-9172-11eb-9191-edd3f9014655.mp4

### DIFF
--- a/packages/app-shell/src/hooks/useOrgSwitcher.js
+++ b/packages/app-shell/src/hooks/useOrgSwitcher.js
@@ -11,7 +11,7 @@ function useOrgSwitcher() {
     )[0];
     const updatedData = {
       account: {
-        ...previousData.account,
+        ...previousData?.account,
         currentOrganization: organizationSelected,
       },
     };


ssvmvss:
Heya @dace, @anafilipadealmeida and @msanroman! 👋 

Sorry for the delay on adding an update on this PR. I started checking this and then other priorities got in the way 😅 I'm going to kickstart a discussion about this on Slack or Threads or Paper, but leaving an update here meanwhile!

The background issue for the error mentioned on this PR is the same one we had before, on other places, were we tried to solve the situation by adding `merge: true` as a merge policy or making sure that the `account` query that was done on each app looked exactly the same than the one that's done in the AppShell. And the cause of all of that is **the way we expose the `billing` data on our API**. 

**How / why is the billing data making the `account` data be null?**

**Because since Apollo has no way of identifying the contents of the `billing` object, Apollo will replace the contents of that object with whatever comes next unless specified otherwise** by adding a `merge: true` policy. 

What is happening on this case in particular is that an app (Analyze, Engage) would requiere some fields for `billing` like:

```
billing {
    canAccessEngagement
}
```

While the AppShell would ask for others, like:

```
billing {
    canAccessEngagement
    canAccessAnalytics
    canAccessPublishing
}
```

**On the cases where the query send by the app would resolve first there would be no problems at all:**

- The query generated by the app resolves and a `billing` object is saved on the cache with `canAccessEngagement` as a field.
- The AppShell query resolves later and, when the response arrives, Apollo has no way of identifying the `billing` object so it replaces it with the new one that contains `canAccessEngagement`, `canAccessAnalytics` and `canAccessPublishing`.
- When the account query generated by the app is read from the cache with `readQuery`, it will resolve OK because all the data that's needed will be there ✅ 
- When the account query generated by the AppShell is read from the cache with `readQuery`, it will resolve OK because all the data that's needed will be there ✅ 

**On the other hand, if the queries resolve on the opposite order, the `account` data will be `null` when it's read from the cache:**

- The query generated by the AppShell resolves and a `billing` object is saved with `canAccessEngagement`, `canAccessAnalytics` and `canAccessPublishing`.
- The query generated by the app resolves and, when the response arrives, Apollo has no way of identifying the `billing` object so it replaces it with the new one that, on this case, it only contains the `canAccessEngagement` field.
- When the account query generated by the app is read from the cache with `readQuery`, it will resolve OK because all the data that's needed will be there ✅ 
- When the account query generated by the AppShell is read from the cache with `readQuery` (as on the case of this PR), it will fail, because it won't be able to resolve the `canAccessAnalytics` or `canAccessPublishing` fields since that data got lost when the object got replaced ❌

**What solutions do we have for this problem?**

**1)** Adding `merge: true` will solve the issue but I personally feel it's risky, prone to bugs or errors and also hard to maintain since you need to make sure your merging policies are the same on all projects, etc.

**2)** Requesting the same data on the billing object on the AppShell + all 4 apps will work, because when objects are replaced all data will be there since it's the same, but it's not only limiting but hard to maintain also. And it makes all benefits of having GraphQL disappear! 🙂 

**3)** Relying on the contents of `useUser` for billing data. If an app needs a field that the AppShell is not requesting, the AppShell needs to be modified (and all 4 apps need to be updated) -> I like this as a temporal solution better, while we implement a better one if we cannot jump on the better solution right away.

**4)** Modifying the API so an `id` (or any other unique) field is provided for the billing data. We could use the organization id for that. This would allow Apollo to identify the object as an identity and we would not have this problem anymore.

As I mentioned above, I'm going to kickstart a discussion for this somewhere else since this discussion involves both frontend and backend and it would be ideal to learn about all of this as a team! 🙂

msanroman:
Thank you so much for your investigation, @ssvmvss 🌮 🌮 !

I agree with your approach here. I think we should default to return and `id`. But I'm curious, since billing is something within `currentOrganization` (and that's the only case we want it in) - **won't we get all the benefits directly if we drop the manual cache writes and start using a proper return payload for `accountSetCurrentOrganization`** ?

If we can get both things done in one go that would be ideal 😄

ssvmvss:
Heya @msanroman! 👋 Great question! 

I think updating the mutation so it returns the full payload will make this issue in particular disappear, but we could still have many other issues (as it happened before on other situations) because of the fact that the `billing` object is always going to get replaced if there are differences between queries.

If we leave mutations / accessing the cache aside, I think we could still have issues when doing just queries, if the fields requested for the billing section vary:
- Component A does a query with X field for `billing` (like the AppShell).
- Component B does a query with Y field for `billing` (like any app).
Every time a new query resolves, the billing object will be replaced. When Apollo tries to solve a query, if any fields are missing, it's going to try to get them from the server. My hunch is that this could lead to some problems depending on how many times the components are rendered, etc.
Also, not sure what could happen if we specify different cache policies than `cache-and-network` or `cache-first`.

We should definitely fix the mutation payload but, if adding an `id` is not too much complication, I would also try to do that to make sure we are safe querying billing data 🙂 

How does this sound to you?

msanroman:
Hey team! (cc: @ssvmvss @dace @anafilipadealmeida)

I wanted to chime in that after all our conversations in the last frontend sync, this week I went ahead and merged bufferapp/core-authentication-service#920 which takes care of this and returns the organization id within the billing node, so this should solve these particular caching woes! 🙏

PR #37, opened 2021-03-30
add trial banner to app shell

gmzjuliana:
Adds the trial banner in the app-shell:
image

### DIFF
--- a/packages/app-shell/src/index.jsx
+++ b/packages/app-shell/src/index.jsx
@@ -15,7 +15,7 @@ import {
 } from './style';
 import { UserContext } from './context/User';
 import { ModalContext } from './context/Modal';
-import useModal  from './hooks/useModal';
+import useModal, { MODALS } from './hooks/useModal';
 import { QUERY_ACCOUNT } from './graphql/account';
 
 /**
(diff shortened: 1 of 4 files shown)


packages/app-shell/src/index.jsx
```diff
@@ -74,6 +89,19 @@ const AppShell = ({
             graphqlConfig={graphqlConfig}
             channels={channels}
           />
+          {isActiveTrial && (
+            <Banner
+              text={trialBannerString}
```
gmzjuliana, line 94: @federicoweber I would've liked to rather send a customHtml but I think since it's react, it's not HTML and I can't off the top of my head remember what we have to do to make it react.. something with renderDOM 🤔  in any case, I can look at it tomorrow if you can't remember either but I wanted to note it!

gmzjuliana: hey @federicoweber I was able to look at this today and what I was looking for was `ReactDOM.renderToString()` https://reactjs.org/docs/react-dom-server.html however, the click handler doesn't fire if I use that 🤔 maybe something we could look at together

federicoweber: Hey @gmzjuliana, yeah that makes sense because, you will be outputting pure HTML, so react won't be working at that stage.

But what is the reason you want to pass down HTML instead of a Component?

gmzjuliana: hey @federicoweber, jejeje I'd expect the click handler to still fire though, clicks are still html 🤔 I tried juuust html and the click handler for a <button onclick="function"> doesn't work either 🤔 I'd love to understand why, even if we don't end up using it

The reason I want to pass HTML is so that the plan name and remaining time can be in bold. Can you think of another way to do that without customHTML? Perhaps the screen shot in the description is misleading, this is what it currently looks like: 
image

federicoweber (changes requested):
Hey @gmzjuliana this is looking great, but let's make sure we are limiting this to OB users before merging it in.

PR #38, opened 2021-03-31
Extract the product name from the url for tracking

dinostheo:
Uses the `getProductPath ` method to extract the product name from the url, instead of the `PRODUCT_TRACKING_KEY` global.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ const ModalTesting = () => (
   <ModalContext.Consumer>
     {modal => (
       <button onClick={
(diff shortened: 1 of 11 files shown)


dinostheo:
@gmzjuliana I combined the product name changed with the CTA changes based on the feedback from the data team, as they were both related to tracking and I could use the result of this PR to the one of the opening modal cta.

dinostheo:
I will merge this one and publish a new patch.
Feel free to review and I can work in potential feedback in a new PR.

PR #45, opened 2021-04-05
add typed errors to planSelector and paymentMethod

gmzjuliana:
- refactors and moves useUpdateSubscriptionPlan hook to `src/hooks` to be used by planSelector and paymentMethod
- returns an error from the hook if the mutation receives a typed error from the backend
- displays the error in both planSelector and paymentMethod


packages/app-shell/src/PlanSelector/components/PlanSelectorContainer.jsx
```diff
@@ -44,9 +45,9 @@ export const PlanSelectorContainer = ({
     return changePlanOptions;
   };
 
-  const [planOptions, setPlanOptions] = useState(
-    filterPlanOptions(changePlanOptions)
-  );
+  const planOptions = filterPlanOptions(changePlanOptions);
```
gmzjuliana, line 48: jejeje look @federicoweber , putting our convo to use, this didn't need to be in state since I don't have to update it and doesn't trigger a re-render on update :)

federicoweber (approved):
Great work on this one @gmzjuliana

PR #46, opened 2021-04-06
Update error redirect to check for unauthenticated code

albennett:
Update graphql error redirect to explicitly check for unauthenticated code

Heya -- I'm back 😊 I reached out to Core around returning a 401 status code and they agreed that it made more sense. However, returning a 400 is part of the Appolo-server standards ([read more here](https://github.com/apollographql/apollo-server/issues/1709)) and they would have to do a workaround to get it to return a 401. 

As long as we're able to discern whether it's an unauthenticated error, then this unblocks us with the missing access pages. Core may still make the 401 status change, but it will not be prioritized before their ANZ launch or other higher priorities. If they do make the change, I've asked them to alert us in the front-end channel first so we can update our logic. 

I'll post about the implications of this change in the front-end channel once it's published.


dace (approved):
LGTM 🚀

ssvmvss:
Ahhhh I'm late for the review (sorry for that!) but great one @albennett! 🙌 I did not know we could identify an unauthorized error that way! 🙂 Great work! 🤗 🌮

PR #47, opened 2021-04-07
Query for organization feature flips and MP subscriptions

dinostheo:
Expands the information under the user object needed for Analyze.


dinostheo:
I will merge and publish this one as it is not breaking the current any current implementation in that way I can base another PR on the product level on the newly created version.

PR #49, opened 2021-04-08
Task/move start trial to hook and expose

gmzjuliana:
This PR moves the start trial mutation to a hook and exposes that hook on the appshell so that we can start trials with one click on any app 🎉


gmzjuliana:
so i merged this and regretted it so i reverted it and opened it again 😬 sorry for the mess y'all!

PR #51, opened 2021-04-08
move start trial mutation to a hook and expose on appshell

gmzjuliana:
This PR moves the start trial mutation to a hook and exposes that hook on the appshell so that we can start trials with one click on any app 🎉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [2.7.5] - 2021-04-12
+- Plan selector QA fixes:
+  - fix spacing and button copy in Start Trial modal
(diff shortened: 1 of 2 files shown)


dinostheo (approved):
Looks good to me @gmzjuliana 
I'm personally curious about the beta versions (2.8.0) 😄. How comes that we started with it? Is a frontend convention?

gmzjuliana:
good question @dinostheo ! I published a beta version so I could use it in my [Analytics PR](https://github.com/bufferapp/buffer-analyze/pull/1044) and have a staging branch to test with, without having to merge this into main in case there was something off. I'll merge this, and use a full version in my Analytics PR before merging that one 😁

dinostheo:
Good one @gmzjuliana. I was doing the same with alpha versions 😂

PR #53, opened 2021-04-12
Default potentially null query data when updating cache

dinostheo:
While I tried to use the org-switcher in production in https://account.buffer.com, I got the following error
Screen Shot 2021-04-12 at 08 08 05

I tried with the analytics app and I got that as well, while trying to impersonate several users and I was able to get the error quite often. When I tried to simulate the error locally it pointed to the apollo client query and this is how I found this line. 

I believe I understand what is happening and I have a theory, but I was not able to think of a proper solution and therefore this "defensive" patch that at least prevents the error and recovers the functionality. 

I have some assumptions though 😅 
In the analytics app we have other graphql queries as well for the account, one for fetching the `canAccessAnalytics` flag, which is not needed anymore and one for fetching the account channels. The account app does the same and actually the org-switcher stops working after visiting the https://account.buffer.com/channels page. 
I don't know how apollo client caching is working but I'm afraid that fetching the same object with various forms might be causing the issue.

### DIFF
--- a/packages/app-shell/src/hooks/useOrgSwitcher.js
+++ b/packages/app-shell/src/hooks/useOrgSwitcher.js
@@ -5,7 +5,7 @@ function useOrgSwitcher() {
   const [setCurrentOrganization] = useMutation(SET_CURRENT_ORGANIZATION);
 
   const updateCache = (organizationId, client) => {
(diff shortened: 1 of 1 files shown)


dinostheo:
I will merge and post PRs for analyze and account. Please do review after it's being merged. It seems to have a deeper cause that we should fix.

PR #57, opened 2021-04-12
Task/style and copy tweaks

gmzjuliana:
in this edition:
- fix borders in plan selector to prevent jumping
- change images for confirmation modal
- start trial and confirmation modal copy changes


dinostheo (approved):
This all looks good @gmzjuliana 
I realize that it would be better to move those copies and image urls to the server side though. If we need to change them soon, it will help us avoid changing the client side again.

gmzjuliana:
hey @dinostheo @federicoweber and I thought about this as we were building and decided not to because these really shouldn't iterate too much. However, I'm keen to re-open the conversation if that's not the case and tackle it in a later iteration. I'll merge this now and publish a new version though just for the sake of time 🙏

PR #61, opened 2021-05-11
Unify Segment Tracking

federicoweber:
To unify events tracking across apps

## How to use
The AppShell, will now be handle `identify`, `group`, and `organizationSwitched` events, and it will do that on every change to the user context to make sure we are sending over the most up to date information.
The rest of the products apps will still be responsible for tracking their own custom events — including page changes, as we have a lot of traits there (ctas, …) that will depend on how the page event has been trigger so we can fully automate it from the appshell.

To do that please use `import { BufferTracker } from '@bufferapp/buffer-tracking-browser-ts'` instead of `window.analytics.track` as this will help with the accuracy of the tracking ([ref.](https://threads.com/34395343006))

The AppShell is also exposing a list of traits getters `packages/app-shell/src/hooks/utils/segmentTraitGetters.js `, please start using those when possible, as they will help with data consistency, and it will make it easier to transition to the APIs Gateway traits getters once we will have those available

Note, as the AppShell now takes care of injecting the Segment tracking script each app will need to expose it's own Segment Write Key as follow
```
    <script type="text/javaScript">
      window.SEGMENT_KEY = "APP_WRITE_KEY";
    </script>
 ```

### CODE REVIEW

File: packages/app-shell/src/hooks/useUserTracker.js [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+      !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addIntegrationMiddleware","setAnonymousId","addDestinationMiddleware"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t,e){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=e};analytics.SNIPPET_VERSION="4.1.0";
```
djfarrelly, line 67: So this will only be injected after this bundle code is downloaded and executed. Will this create a race condition perhaps with other tracking that is used @federicoweber? Can we still recommend that this is only a fallback and it's ideal to have this snippet on each app's `index.html`?

federicoweber: Good one @djfarrelly , it should not, at least at the moment, but could happen 🤔.
👍 on keeping the injection as a fallback, I've already made it conditional anyway to edge on safety 🙂.

PR #65, opened 2021-05-18
remove the MPBilling trial field that was preventing the query from resolving

federicoweber:
This is a temporary solution, we will need to fix this on the BE, making sure the `trial` is set to `null`, when missing for `MPUsers` instead of getting omitted.


packages/app-shell/src/index.jsx
```diff
@@ -42,6 +42,11 @@ const AppShell = ({
       }
     : {};
   const { data, loading, error } = useQuery(QUERY_ACCOUNT, graphqlConfig);
+  window.__userData = {
```
federicoweber, line 45: this is to simplify debugging from other apps

PR #69, opened 2021-06-04
test staging env

federicoweber:


kiriappeee:
heyya @federicoweber . I think you are on the right track here for getting this deployed. I think what you want to try first though is to build the image locally and see if it runs. When I tried to do that, I got an error where it said that the modules could not be imported. The error I get is:


```
Starting the development server...

Failed to compile.

./src/App.js
Module not found: Can't resolve '@bufferapp/app-shell' in '/app/packages/app/src'
Compiling...
Failed to compile.

./src/App.js
Module not found: Can't resolve '@bufferapp/app-shell' in '/app/packages/app/src'
```

I ran the following: 

```
docker build -t appshell:test .
docker run --rm appshell:test
```

to test. Once we have this step sorted, we can move forward with more config very quickly 😃 .

federicoweber:
Thanks for looking into this @kiriappeee, with the last changes we are now building and serving the production version of the apps.
And I can see the container it's running properly, but this branch is not triggering the deployment anymore 😕

package.json
```diff
@@ -12,10 +12,15 @@
...
   "scripts": {
-    "watch": "npm-run-all --parallel app app-shell",
-    "app-shell": "lerna exec --scope @bufferapp/app-shell -- yarn watch",
     "app": "lerna exec --scope app -- yarn watch",
+    "app-shell": "lerna exec --scope @bufferapp/app-shell -- yarn watch",
+    "build": "npm-run-all --parallel build:app build:app-shell",
+    "build:app": "lerna exec --scope app -- yarn build",
+    "build:app-shell": "lerna exec --scope @bufferapp/app-shell -- yarn build",
+    "serve": "serve -l 3000 ./packages/app/build",
+    "serve:ssl": "serve -l 3000 --ssl-cert ../reverseproxy/certs/local.buffer.com.crt --ssl-key ../reverseproxy/certs/local.buffer.com.key ./packages/app/build",
```
federicoweber, line 21: @kiriappeee at the moment I'm using the `serve` script to run the app but we will need to use `serve:ssl` to access serve the secure page and access the APIs Gateway.
In the script I'm using the local certificates, but not sure how to change it to run in the staging container 🙂 .

helm/testenv.yaml
```diff
@@ -0,0 +1,36 @@
...
+ingress:
+  enabled: true
+  path: /
+  annotations:
+    kubernetes.io/ingress.class: nginx
+image:
+  repository: bufferapp/app-shell
+resources:
+  limits:
+    memory: 32Mi
```
kiriappeee, line 30:
```suggestion
```

kiriappeee: Sorry I didn't add more description here. Wasn't sure I'd be able to commit the suggestion if I did. I ran into a problem when I deployed it today in the evening. The pod crashed instantly and did not start up. I checked and it said OOM killed. My guess is that the resources aren't enough since it's just 32 Mi that's been added as a limit. So I've removed the limits to test that assumption for now.

kiriappeee:
Heyya @federicoweber . I've begun testing this. Just to clarify on how we can deploy this until we've added this to the pipeline:

`@bufferbot servicedeploy task/staging to app-shell` will deploy the service.

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM node:12-alpine
+WORKDIR /app
+COPY . /app
+RUN yarn install
+RUN yarn add -W serve
+
+CMD yarn build && yarn serve
```
kiriappeee, line 7: @federicoweber  I think this should just be `yarn serve`? We shouldn't be running `yarn build` as a startup command. The assets should be built already at this point I believe.

kiriappeee (changes requested):

Dockerfile
kiriappeee: The error I'm seeing in the container now is:

```
yarn run v1.22.5
$ npm-run-all --parallel build:app build:app-shell
Killed
```

This makes sense since the build is probably taking a lot of memory. IF this is a must, then we'd need to bump the resource request numbers to ensure that the container is started with more resources allocated to it by default.

In any case, I think starting a container with just 16 Mi might be causing trouble if the pod is placed into a machine without more resources to expand. For context, [publish starts with 80 Mi for memory and 150m for CPU](https://github.com/bufferapp/buffer-publish/blob/main/helm/buffer-publish.yaml#L22-L24)

kiriappeee, line 7:
```suggestion
RUN yarn add -W serve && yarn build
CMD yarn serve
```

package.json
kiriappeee: @federicoweber I think we can tackle this as soon as we get this up and running for the first time. My hunch is that this might not be an issue. Our apps run behind HTTPS proxies by default. But if the app needs to have an SSL certificate of its own embedded and it needs to run as an HTTPS app (i.e., not an HTTP app sitting behind an HTTPS proxy), then we'll need to do some additional work here which we can figure out soon.

federicoweber: ah good one, I think the proxy will be enough. Will change the config so it's not forcing certificates on staging

Dockerfile
federicoweber: Ah that makes sense. What do you think could be a good setting? 32 Mi? I think we won't need much CPU once it's built.

federicoweber:
This is looking great @kiriappeee 
image

Thanks for the help 🌮 🙌 .

Going to merge this is. Do you think we can enable automatic deployment for PRs?

PR #72, opened 2021-06-15
Fix App Shell Distribution

federicoweber:
This updates the App Shell to fix the distribution and the isolation problems.
 
With this changes the AppShell is not wrapping the Product Apps anymore, is distributed as a self contained JS file, and should implemented as follow:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <!-- This style reset is required for all Apps to make sure we are not seeing any glitch on load -->
    <style type="text/css">
      #navigator {
        min-height: 57px;
      }

      #root {
        background-color: #F5F5F5;
        display: flex;
        flex-direction: column;
        height: 100%;
        margin: 0;
        overflow: auto;
      }
    </style>
  </head>
  <body>
    <!-- This is the div used to render the Navigator -->
    <div id="navigator"></div>
    <!-- This is the div used to render the App -->
    <div id="root"></div>

    <!-- Define the URL for the APIs Gateway used by the App Shell -->
    <script type="text/javaScript">
      window.API_GATEWAY_URL= "https://graph.local.buffer.com";
    </script>
    <script src="https://components.buffer.com/navigator.js" async></script>
  </body>
</html>

```

---
For more details please read the [RFC doc](https://paper.dropbox.com/doc/RFC-App-Shell-architecture-evolution--BNSCEwIcEMWLX_0ZPp0Ce9QHAg-EHu558jTk6hTaedWa010S)


packages/app-shell/src/appShell.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import exposeAppShell from './index';
+
+exposeAppShell()
```
federicoweber, line 4: This is the entry point of the ` AppShell package.

packages/app-shell/src/index.jsx
```diff
@@ -224,10 +221,40 @@ AppShell.defaultProps = {
...
+    }
+  }
+
+  ReactDOM.render(
+    <React.StrictMode>
+      <ApolloProvider client={client}>
+        <AppShell />
+      </ApolloProvider>
+    </React.StrictMode>,
+    document.getElementById('appShell')
```
federicoweber, line 257: To keep things isolated we are rendering the AppShell to its own div.

packages/app-shell/src/index.jsx
```diff
@@ -224,10 +221,40 @@ AppShell.defaultProps = {
...
+        window.dispatchEvent(event)
+      }
+    }
+  }
+
+  ReactDOM.render(
+    <React.StrictMode>
+      <ApolloProvider client={client}>
+        <AppShell />
+      </ApolloProvider>
```
federicoweber, line 255: As the AppShell it's now self-contained it need its own `ApolloProvider`

packages/app-shell/src/index.jsx
```diff
@@ -224,10 +221,40 @@ AppShell.defaultProps = {
...
+    modal: {
+      MODALS,
+      openModal(modal, data) {
+        const event = new CustomEvent(MODAL_EVENT, {
+          detail: { modal, data }
+        })
+        window.dispatchEvent(event)
+      }
+    }
+  }
```
federicoweber, line 249: We are exposing a global object to allow infra components communication.
For example this is now used to open modals from outside of the AppShell, as the `ModalContext` is not longer available to the Apps.

packages/app/src/App.js
```diff
@@ -1,47 +1,20 @@
...
 
-const AnotherComponentRenderingUserData = () => {
-  const user = useUser();
-  const switchOrganization = useOrgSwitcher()
+const ModalTesting = () => {
+  const { MODALS, openModal } = window.appshell.modal;
 
+  return (<>
+      <h2>Render Plan Selector</h2>
+      <button onClick={() => {openModal(MODALS.planSelector, { cta: 'renderModal', ctaButton: 'renderModal', isUpgradeIntent: false })}}>Render Modal</button>
```
federicoweber, line 8: Quick example on how to open a modal from an App

packages/app-shell/src/hooks/useModal.js
```diff
@@ -20,6 +22,13 @@ function useModal() {
     }
   }, [])
 
+  useEffect(() => {
+    window.addEventListener(MODAL_EVENT, (e) => {
+      const { modal:key, data } = e.detail
+      openModal(key, data)
+    })
+  })
+
```
federicoweber, line 31: Each component that is expected to handle communication with the resto of the apps should expose it's own `EVENT_KEY`, and define it's own listener.

packages/app/public/index.html
```diff
@@ -42,5 +43,6 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script src="https://appshell.local.buffer.com:8085/main.js" async></script>
```
federicoweber, line 46: this will be driven from the ENV to point to the proper version.

federicoweber (commented):
Hey @anafilipadealmeida, @djfarrelly, @hamstu, @msanroman, @ssvmvss, this is an early draft but it's working locally and it will be great if you could take a look.
Looking forward to your feedback.

Thanks!

packages/app-shell/src/hooks/useModal.js
```diff
@@ -7,6 +7,17 @@ export const MODALS = {
...
+export const ACTIONS = {
+  openModal(modalKey, modalData = null){
+    const event = new CustomEvent(EVENT_KEY, {
+      detail: { modal: modalKey, data: modalData }
+    })
+    window.dispatchEvent(event)
+  }
+}
+
```
federicoweber, line 20: if a component needs to communicate with the Product Apps it can do so by exposing an `EVENT_KEY` and a series of actions it supports. Those will then made available to the global `window.appshell` object

packages/app/src/App.js
```diff
@@ -1,47 +1,20 @@
...
 
-const AnotherComponentRenderingUserData = () => {
-  const user = useUser();
-  const switchOrganization = useOrgSwitcher()
+const ModalTesting = () => {
+  const { MODALS, actions:modalActions } = window?.appshell?.modal || {};
 
+  return (<>
+      <h2>Render Plan Selector</h2>
+      <button onClick={() => {modalActions.openModal(MODALS.planSelector, { cta: 'renderModal', ctaButton: 'renderModal', isUpgradeIntent: false })}}>Render Modal</button>
```
federicoweber, line 8: simple example on how to open a modal from a Product App

packages/app/src/App.js
```diff
@@ -1,47 +1,35 @@
...
+
+  useEffect(() => {
+    const { eventKey:organizationEventKey } = window?.appshell?.organization || {};
+
+    window.addEventListener(organizationEventKey, handleOrgSwitch)
+
+    return function cleanup() {
+      window.removeEventListener(organizationEventKey, handleOrgSwitch)
+    };
+  }, [])
```
federicoweber, line 26: example on how to listen to dispatched events

packages/app-shell/src/appShell.js
```diff
@@ -0,0 +1,36 @@
...
+      background-color: ${grayLighter};
+      display: flex;
+      flex-direction: column;
+      height: 100%;
+      margin: 0;
+      overflow: auto;
+    }
+  `
+  document.head.append(style);
+}
```
federicoweber, line 31: This used to be provided by the AppShell itself, but given that we are not wrapping the app anymore we need to inject the style to avoid breaking the layout.

It won't be needed anymore once we have a true app shell that provides the layout, and compose the apps based on the url.

packages/app-shell/src/NavBar/index.jsx
```diff
@@ -288,6 +283,42 @@ const NavBar = React.memo((props) => {
...
+    {
+      id: 'Quick Help',
+      title: 'Quick Help',
+      onItemClick: () => {
+        if (window.zE) {
+          window.zE('webWidget', 'show');
+          window.zE('webWidget', 'open');
+        }
+      },
+    },
```
federicoweber, line 304: At the moment this is just going to open the Zendesk Widget if that's present in the page, but I'm thinking we might be able to also consolidate the actual Zendesk integration in here (Maybe reusing the custom hook from [Engage](https://github.com/bufferapp/buffer-engage-client/blob/82b49e29e7adc6c215339d3d32dd6883723961e4/src/shared/hooks/useZendeskWidget/useZendeskWidget.js#L2)). 

@mickmahady would love your opinion on this one 🙂 .

mickmahady: Hey @federicoweber! Thanks for the ping here! I'm all in favour of consolidating the Zendesk integration and having one implementation live in the app-shell rather than 4 separate products. I think the Engage hook is probably the simplest form of the integration. We have more complex needs for the other products, such as suggesting different FAQ articles in the widget when a user visits a different page/tab in the product. Examples:
- [Publishing](https://github.com/bufferapp/buffer-publish/blob/main/packages/thirdParty/zendesk-widget/index.js#L1-L35)
- [Analytics](https://github.com/bufferapp/buffer-analyze/blob/main/packages/zendesk-widget/index.js#L1-L29)
- [Account](https://github.com/bufferapp/buffer-account/blob/main/packages/zendesk-widget/index.js#L1-L23)

We'd need to make sure that functionality is still supported after moving it into the App Shell! Of course, this is now more complex than you first thought, and I don't expect you to take this on. If Zendesk isn't a blocker for this PR, please feel free to leave it and I can add it to my team's backlog for an upcoming cycle :grin: 

Kinda related, we have a task to add a "group" call in Segment for each product, which is called when we identify a user and when they switch organizations in the App Shell. I'm wondering if this can now be done in this code base, rather than each individual code base? Dan added [this group call](https://github.com/bufferapp/buffer-publish/blob/main/packages/analytics-middleware/middleware.js#L32-L39) for Publish but we want to add it for all products and also not just when the org switcher is used. We want the group call to be triggered even when we identify the user, so any actions they take are always associated with the currently selected organization.

federicoweber: Thanks for the extra context @mickmahady. 💯on doing the Zendesk consolidation in a follow-up PR. Do you think that is something Internal Tools could work on?

As for the group call, you are also spot-on in doing this from one codebase, in fact, we are already doing that 🙂 . I've added it in the AppShell a few weeks ago in this PR https://github.com/bufferapp/app-shell/pull/61 😉 .

packages/app-shell/uploadStatic.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+apk add curl
+UPLOADER="https://github.com/bufferapp/buffer-static-upload/releases/download/0.3.0/buffer-static-upload-`uname -s`"
+curl -L $UPLOADER > ./buffer-static-upload
+chmod +x ./buffer-static-upload
+
```
kiriappeee, line 6:
```suggestion
ls .
```

packages/app-shell/uploadStatic.sh
```diff
@@ -0,0 +1,10 @@
...
+cd build
```
kiriappeee, line 7:
```suggestion
cd build
ls .
ls ..
```

packages/app-shell/src/NavBar/index.jsx
mickmahady: > Do you think that is something Internal Tools could work on?

Absolutely! I'll add it to our backlog and chat with Amanda about putting it in our next cycle. It should be a good opportunity for one of our new teammates to get familiar with working in the App Shell. You can leave the Zendesk consolidation to us now :grin: 

> I've added it in the AppShell a few weeks ago in this PR #61.

Awesome! I'll report back to our team that we don't need to take that on. Thanks so much, @federicoweber :raised_hands:

packages/app-shell/src/hooks/utils/eventDispatcher.js
```diff
@@ -0,0 +1,9 @@
+function eventDispatcher(eventKey, detail) {
+  const event = new CustomEvent(eventKey, {
+    detail,
+  })
+
+  window.dispatchEvent(event)
```
msanroman, line 6: I love this!

Not to derail the implementation, but `new CustomEvent` does not work on IE - do we need to support it?

msanroman (commented):
Great work @federicoweber - I'd love if we can change the layout stuff to be directly on the HTML of each product to prevent layout shift, but other than that, this looks amazing ✨ 

Very excited for this new stage of the AppShell/NavBar 😉

packages/app-shell/src/appShell.js
federicoweber: Good one @msanroman, I was also playing with a few options after yesterday FE-WG call, we can move this to a css file that is served similarly from a static URL. If we import that in the header it should prevent shifts.
But you are right, it's a small amount of css it's fine to duplicate this for now.

P.s. We still risk having some shifts if the navigation is rendering a banner, as that will push the `#root` down

packages/app-shell/src/hooks/utils/eventDispatcher.js
federicoweber: Good one, I was under the impression we are not supporting IE anymore.

packages/app-shell/src/appShell.js
```diff
@@ -0,0 +1,36 @@
...
+window.appshell = {
+  eventKeys: {
+    ORGANIZATION_EVENT_KEY,
+    MODAL_EVENT_KEY,
+  },
+  actions: {
+    ...MODAL_ACTIONS,
+  },
+  MODALS,
+}
```
federicoweber, line 16: I've simplified this object to avoid excessive nesting, it will be a bit leaner if we start adding more actions and events

packages/app-shell/src/NavBar/index.jsx
```diff
@@ -288,6 +291,70 @@ const NavBar = React.memo((props) => {
...
   }
 
+  const menuItems = [
+    // This is only needed for Publish
+    (window.location.href.match(/publish/)) ? {
+      id: 'preferences',
+      title: 'My Preferences',
+      icon: <GearIcon color={gray} />,
+      onItemClick: () => {
+        window.location.pathname  = '/preferences/general'
```
federicoweber, line 301: Unfortunately this will force a reload instead of a smooth page change. If we want to make it smoother we will have to expose the `dispatch(navigate(...)` from Publish

packages/app-shell/webpack.config.dev.js
```diff
@@ -0,0 +1,27 @@
...
+    compress: true,
+    disableHostCheck: true,
+    https: true,
+    host: '0.0.0.0',
+    port: '8085',
+    hot: true,
+  },
+  plugins: [
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.DefinePlugin({
```
hamstu, line 21: 💅🏻 nit; the environment plugin makes this a bit cleaner to write: (no need for stringify) https://github.com/bufferapp/buffer-publish/blob/4fe9e64a961ba2f2d96e8c451e1822d3177078eb/webpack.config.dev.js#L55-L60

packages/app-shell/src/index.jsx
```diff
@@ -224,10 +202,27 @@ AppShell.defaultProps = {
...
 export { default as useOrgSwitcher } from './hooks/useOrgSwitcher';
 export { MODALS } from './hooks/useModal';
+
+export default () => {
+  const client = new ApolloClient({
+    cache: new InMemoryCache(),
+    link: new HttpLink({
+      uri: GRAPHQL_API,
+      credentials: 'include',
+    }),
```
hamstu, line 217: we should pass an x-buffer-client-id header in here, see https://github.com/bufferapp/buffer-account/blob/c40639840f145794fbf5e045c4599d6228b4922a/packages/web/src/apollo/providers/APIGatewayProvider/APIGatewayProvider.jsx#L21

packages/app-shell/webpack.config.dev.js
```diff
@@ -0,0 +1,27 @@
...
+const mode = 'development';
+
+const merged = merge(common, {
+  mode,
+  devtool: 'inline-source-map',
+  devServer: {
+    contentBase: './dist',
+    compress: true,
+    disableHostCheck: true,
+    https: true,
```
hamstu, line 14: can get proper https working (no warnings) with this: https://github.com/bufferapp/buffer-publish/blob/4fe9e64a961ba2f2d96e8c451e1822d3177078eb/webpack.config.dev.js#L37-L44

packages/app/package.json
```diff
@@ -17,10 +17,10 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "REACT_APP_PATH='/static/js/' REACT_APP_FILE_NAME='navigator.js' react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "watch": "HTTPS=true HOST='0.0.0.0' react-scripts start"
+    "watch": "HTTPS=true HOST='0.0.0.0' REACT_APP_PATH='https://appshell.local.buffer.com:8085/' REACT_APP_FILE_NAME='main.js' react-scripts start"
```
hamstu, line 23: more of an observation but we're getting to the point that I think we need a notion page or something to track all the 'ports in use' for local dev 😅 — each app I think uses diff. ones now, but it could be helpful to know which

hamstu (approved):

packages/app-shell/src/NavBar/index.jsx
djfarrelly: @federicoweber what about `window.history.pushState()`? I wonder if React Router listens to this and it works automatically? 🤔 

Alternatively, there could be a way for the app to register itself w/ the appshell/appnav, e.g. `window.appshell.init({ router })

.github/workflows/upload.yml
```diff
@@ -0,0 +1,21 @@
+name: Upload Navigator to components.buffer.com
+# TODO limit this to the main branch
+on: [ push ]
```
djfarrelly, line 3: @federicoweber should we change this to run only on `main` when this work is done? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on

federicoweber: yep, I was running it on all branches for test only :)

PR #73, opened 2021-06-30
show channels in org switcher

federicoweber:
To show the channels list in the org switcher.
image


hamstu (commented):
Thanks @federicoweber! 💯

anafilipadealmeida:
Happy to see we're finally able to do this! 🎉 No more duplicated code!

PR #78, opened 2021-07-09
Add GraphQL business logic to handle OB Migration

federicoweber:


packages/app-shell/src/common/graphql/billing.js
```diff
@@ -65,3 +65,59 @@ export const START_TRIAL = gql`
     }
   }
 `
+export const MIGRATE_TO_OB = gql`
+  mutation BillingMigrateToOneBufferResponseMutation(
+    $billingMigrateToOneBufferOrganizationId: String!
```
federicoweber, line 70: hey @dinostheo and @esclapes I see this hasn't been exposed to the APIs Gateway yet. Is there any blocker?

packages/app-shell/src/common/graphql/billing.js
```diff
@@ -65,3 +65,59 @@ export const START_TRIAL = gql`
...
+                details
+                intervalBasePrice
+                intervalUnit
+              }
+              isRecommended
+              downgradedMessage
+            }
+          }
+        }
+      }
```
federicoweber, line 119: This is a bit redundant, but it's used to properly override the data from the account query, we should be able to clean this up once we return an `ID` for the `billing` object.

dinostheo: @federicoweber `Billing` does have an `id`. It is the `organizationID`, but it should work well for the cache. 

Related to schema IDs:
As part of the working groups I started research on this subject as well and I have documented the finding so far in [this paper](https://paper.dropbox.com/doc/RFC-Apollo-client-InMemoryCache-and-Object-ids--BOPpXfO~Xh7En9bRBfeJ2cU9Ag-bN3qCCI3Q1lFErDcWdB9O)

packages/app-shell/src/common/graphql/billing.js
dinostheo: @federicoweber you are right. That's my fault. 
I forgot to change the API gateway when I merged the migration PR. I will do it on Monday morning.

esclapes (approved):
This looks good to me @federicoweber :taco: :+1: 

It's great to see an example of typed errors, I pinged Maya and Joe below to discuss the pattern because that's part of the backend wg current agenda, but that should not block this PR :rocket:

esclapes (changes requested):
Hey @federicoweber I was chatting with Phil about this and we thought it may be a good idea to make two changes before merge. That way we can use this repo to explore/promote new best practices :

* let's use a Fragment to avoid the repetition on the queried fields
* let's figure out a way to use the underlying `__typename` instead of accessing a property like `userFriendlyMessage` to check if it's an error.

Thanks! :raised_hands:

packages/app-shell/src/common/graphql/billing.js
federicoweber: thanks @dinostheo

packages/app-shell/src/common/graphql/billing.js
federicoweber: Good one @esclapes, you are spot on, that's why I mentioned we should use IDs instead. With IDs, we shouldn't request a refetch either, as Apollo Client we be handling this for us.

@dinostheo that's great news, will change this to use the ID then. Thanks!

packages/app-shell/src/common/hooks/useMigrateToOB.js
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    console.log(data);
+    if (mutationError) {
+      setError(mutationError);
+      setProcessing(false);
+    } else if (data?.billingMigrateToOneBuffer.userFriendlyMessage) {
+      setError({ message: data?.billingMigrateToOneBuffer.userFriendlyMessage });
+      setProcessing(false);
+    }
+  }, [data, mutationError]);
```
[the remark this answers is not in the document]
federicoweber: Hey @esclapes, I get your point, but I actually prefer to use the field in here instead of the type, as what we really care for is to show the `userFriendlyMessage` if that's available, so this feels a bit more explicit. As for the `mutationError` that could happen for any reason, including lost connection, so we will always need to check for that as well.

packages/app-shell/src/common/graphql/billing.js
```diff
@@ -65,3 +65,19 @@ export const START_TRIAL = gql`
...
+export const MIGRATE_TO_OB = gql`
+  mutation BillingMigrateToOneBufferResponseMutation(
+    $billingMigrateToOneBufferOrganizationId: String!
+  ) {
+    billingMigrateToOneBuffer(
+      organizationId: $billingMigrateToOneBufferOrganizationId
+    ) {
+      ... on BillingMigrateToOneBufferResponse {
+        billing {
+          id
```
esclapes, line 77: Hey @federicoweber, I am not sure I understand this change.

When the mutation executes, most `billing` details will change. If we only receive back the `id`, how are the new value going to be updated in the cache?

My hunch is that by only requesting the `id` the billing fields for the organization will be out of date. I may be missing something, but if we create [BillingFields fragment](https://www.apollographql.com/docs/react/data/fragments/) we could define the list of fields the AppShell cares about and use the fragment in all 3 places:
* `currentOrganization.billing`
* `organizations.billing`
* `billingMigrateToOneBuffer`

If the fragment includes the `id`, I would expect the cache invalidation to work fine without need to re-fetch. I may be missing something about how caching works, though, so.

packages/app-shell/src/common/hooks/useMigrateToOB.js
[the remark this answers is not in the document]
esclapes: > I get your point, but I actually prefer to use the field in here instead of the type, as what we really care for is to show the userFriendlyMessage if that's available

Sound good to me, @federicoweber :+1:  This won't be hard to iterate on as we introduce new types of errors. cc @philippemiguet

packages/app-shell/src/common/graphql/billing.js
federicoweber: You are right @esclapes I was under the impression that getting back the ID would have been enough to trigger a cache invalidation, but it won't work by itself unless the rest of the fields are present.
I like your approach, we could also solve this by triggering a re-fetch,

```
const [migrateToOB, { data, error: mutationError }] = useMutation(
  MIGRATE_TO_OB,
  {
    refetchQueries: [{ query: QUERY_ACCOUNT }],
  }
);

```
but the Fragment is more elegant and save us a request.

esclapes (approved):
LGTM @federicoweber! Nice example of Fragments that we have now :heart_eyes:

philippemiguet:
Great use of Fragment there! 🙌

PR #79, opened 2021-07-14
 expose setCurrentOrganization action

federicoweber:
This expose the `setCurrentOrganization` action so that product apps can trigger the organization change outside of the Navigator.
This is needed to allow Publishing to change the org based on the url.

### DIFF
--- a/packages/app-shell/src/common/hooks/useModal.js
+++ b/packages/app-shell/src/common/hooks/useModal.js
@@ -48,12 +48,18 @@ function useModal() {
   }, [])
 
   // Open modal from events
+  function handleOpenModal({ detail }) {
+    const { modal:modalKey, data:modalData } = detail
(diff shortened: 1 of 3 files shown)


packages/app-shell/src/common/hooks/useModal.js
```diff
@@ -48,12 +48,18 @@ function useModal() {
...
-      const { modal:modalKey, data:modalData } = e.detail
-      openModal(modalKey, modalData)
-    })
-  })
+    window.addEventListener(EVENT_KEY, handleOpenModal)
+
+    return function cleanup() {
+      window.removeEventListener(EVENT_KEY, handleOpenModal)
+    }
+  }, [])
```
federicoweber, line 62: clean up the event handler 😉

PR #80, opened 2021-07-19
fix local for OSx

federicoweber:
fix an issue with local dev env in `OSx` caused by the case difference between file and folder.


gisete:
Alt Text

PR #82, opened 2021-07-21
fix trial expired rendering

federicoweber:


packages/app-shell/src/components/Modal/modals/TrialExpired/index.jsx
```diff
@@ -91,7 +91,8 @@ export const Modal = ({
   const imageUrl = 'https://buffer-ui.s3.amazonaws.com/Confirmation+Illustration.png';
   const description = `Your trial is over and you are back to free features. Upgrade to get the power restored.`;
   const planId = user?.currentOrganization?.billing?.subscription?.plan?.id
-  const planDetails = user?.currentOrganization?.billing?.changePlanOptions.find(o => o.planId === planId)?.summary.details
+  const changePlanOptions = user?.currentOrganization?.billing?.changePlanOptions || []
```
djfarrelly, line 94: @federicoweber - thinking about catching this earlier - could we add automated unit tests for some of these things? We could extract the logic to a `getPlanDetails(user: User)` type function that could even have a default or a try catch.

federicoweber: yep you are right, we should have this under Unit Tests

PR #93, opened 2021-08-16
Grow 195 ux navbar

daisymarie128:
Link to ticket: https://buffer.atlassian.net/browse/GROW-195

This PR updates the styling of the app navbar. You can find the design specs in the ticket linked.

### 🤹‍♂️Main things happening in the PR:
- New **_"Apps"_** dropdown menu added.
- Tweaks to hover states and style changes
- New 1200px breakpoint was added to reduce some padding in navbar.
- Moved out dropdown constants to a utils file to try and move some code out from the `.../components/NavBar/index` file. I chose to move this out to make it easier to find information in this file 

### Screenshots:
Screen Shot 2021-08-16 at 3 52 16 pm

Hover state:
Screen Shot 2021-08-16 at 3 52 25 pm

Active state: 
Screen Shot 2021-08-16 at 3 53 59 pm

Apps dropdown:
Screen Shot 2021-08-17 at 7 40 40 am


packages/app-shell/src/components/NavBar/components/NavBarMenu/NavBarMenu.jsx
```diff
@@ -53,6 +53,7 @@ NavBarMenu.propTypes = {
   user: PropTypes.shape({
     name: PropTypes.string,
     email: PropTypes.string.isRequired,
+    isImpersonation: PropTypes.bool.isRequired,
```
daisymarie128, line 56: 📓Note: I added this here to fix an eslint error stating this was never defined. (Wasn't sure how to check if this property is always passed through in the data of `user`)

packages/app-shell/src/components/NavBar/components/NavBarProducts/NavBarProducts.jsx
```diff
@@ -2,51 +2,57 @@ import React from 'react';
...
+import { grayDark, green, white, blue } from '@bufferapp/ui/style/colors';
 
 import {
   fontWeightMedium,
+  fontWeightBold,
   fontFamily,
   fontSizeSmall,
 } from '@bufferapp/ui/style/fonts';
 
+const lighterGreyHoverState = 'rgba(218, 221, 238, 0.25);';
```
daisymarie128, line 14: 📓 Note: This const has been declared in 3~ different spots. I wasn't sure if there was a better way to share this variable around or where to put it as most colors come from `@buffer/ui/style` 🤔 if you have any suggestions let me know 😄

packages/app-shell/src/components/NavBar/components/NavBarMenu/NavBarMenu.jsx
federicoweber: Good one 🤔, theoretically this property is always present but I'm not sure we want to make it as a requirement as the NavBar will be able to properly render even if it's missing.
I would suggest setting it as an optional parameter with the default set to `false`

packages/app-shell/src/components/NavBar/components/NavBarMenu/style.js
```diff
@@ -14,6 +13,8 @@ import {
   fontFamily,
 } from '@bufferapp/ui/style/fonts';
 
+const lighterGreyHoverState = 'rgba(218, 221, 238, 0.25);'
```
federicoweber, line 16: Hey @daisymarie128 I think for colors we want to use the ones defined in the [UI guidelines](https://www.notion.so/buffer/cab977d026c748009c4d9c70b08a21f5?v=6fa298b9773b40fb8636c38c055fa8a3&p=80f6c8b6ec6d45edab660b5c34fc9424) as much as possible. 
Any specific reason we are changing this one?

packages/app-shell/src/components/NavBar/components/NavBarProducts/NavBarProducts.jsx
federicoweber: Great intuition there. Feels like the colors exports in UI will be a better place for this.
But I will also double-check with the Desing team if we want to add an extra one 🙂.

packages/app-shell/src/components/NavBar/index.jsx
```diff
@@ -42,6 +32,12 @@ import { MODALS } from '../../common/hooks/useModal';
 import useOrgSwitcher from '../../common/hooks/useOrgSwitcher';
 import { isFreePlan } from '../../common/hooks/utils/segmentTraitGetters';
 
+import NavBarDropdown from './components/NavBarDropdown';
+import {
+  getHelpDropdownItems,
+  getAppsDropdownItems,
+} from './utils/dropdown-items';
```
federicoweber, line 39: Great call abstracting those 💯.

federicoweber (changes requested):
Hey @daisymarie128 thanks for working on this. 
I left a couple of small comments I would like to address before shipping this but overall it's looking solid 🌮.

packages/app-shell/src/components/NavBar/components/NavBarProducts/NavBarProducts.jsx
daisymarie128: @federicoweber I've removed these styles now and replaced with a gray from buffer/ui 😊

federicoweber (approved):
Looks great, thanks @daisymarie128 
ship it

PR #101, opened 2021-08-18
Hide migration modal if Pendo guide is visible

gisete:
If there is a Pendo guide in place on the page it overlaps the OB migration modal. Talking to Tom and James about it we decided it would be great to only display the migration modal if there are no Pendo guides visible.

Screen Recording 2021-08-18 at 03 53 42 PM

I'm using a function in Pendo's script called `pendo.isGuideShown()` to check if any Pendo guides are up on the page. I don't have a ton of context about how Pendo works and I couldn't find any documentation on all the functions available so I'm not sure if this covers all cases.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,10 @@ module.exports = {
   plugins: ['react', 'testing-library', 'jest-dom'],
   rules: {
     "import/prefer-default-export": "off",
+    "spaced-comment": "off",
+    "no-unneeded-ternary": "off",
+    "import/no-named-as-default": "off",
+    "react/require-default-props": "off"
   },
   ignorePatterns: ['lib/*'], // Stop ESLint complaining when looking at transpiled lib
 };
--- a/packages/app-shell/src/components/Modal/index.jsx
+++ b/packages/app-shell/src/components/Modal/index.jsx
@@ -95,19 +95,12 @@ const Modal = ({ modal, openModal }) => {
 
       if(window.pendo && !window.pendo.isGuideShown()) {
         openModal(MODALS.paidMigration);
-        return;
       }
       else {
         if (!window.pendo) {
           openModal(MODALS.paidMigration);
         }
       }
-      // if pendo exists 
-        // then check if pendo window is open
-            // if yes then dont show modal
-            // if no show modal
-      // if pendo doesnt exists
-        // show modal 
     }
   }, [user.loading]);
 
--- a/packages/app-shell/src/components/Modal/index.jsx
+++ b/packages/app-shell/src/components/Modal/index.jsx
@@ -17,7 +17,6 @@ import Success from './modals/PaidMigration/Success';
 import StickyModal from './modals/StickyModal';
 
 const ModalContent = ({ modal, closeAction }) => {
-  console.log("here: " + typeof modal);
   switch (modal) {
     case MODALS.paymentMethod:
       return (
@@ -79,7 +78,7 @@ const ModalContent = ({ modal, closeAction }) => {
 };
 
 ModalContent.propTypes = {
-  modal: PropTypes.object.isRequired,
+  modal: PropTypes.objectOf(PropTypes.object).isRequired,
   closeAction: PropTypes.func.isRequired,
 };
 
@@ -95,7 +94,7 @@ const Modal = ({ modal, openModal }) => {
       openModal(MODALS.trialExpired)
     }
 
-    // Check if Pendo loads on the page - we don't want to show the OB Migration modal if there is a Pendo guide already visible
+    //Check if Pendo loads on the page - we don't want to show the OB Migration modal if there is a Pendo guide already visible
     const isPendoModalVisible = window.pendo && !window.pendo.isGuideShown() || !window.pendo ? false : true;
 
     //Migrate to OB modal
@@ -115,7 +114,7 @@ const Modal = ({ modal, openModal }) => {
 };
 
 Modal.propTypes = {
-  modal: PropTypes.object,
+  modal: PropTypes.objectOf(PropTypes.object),
   openModal: PropTypes.func.isRequired,
 };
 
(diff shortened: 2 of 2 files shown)


.eslintrc.js
```diff
@@ -14,6 +14,10 @@ module.exports = {
   plugins: ['react', 'testing-library', 'jest-dom'],
   rules: {
     "import/prefer-default-export": "off",
+    "spaced-comment": "off",
```
dinostheo, line 17: @gisete I 'm not sure if these changes were committed accidentally or if they are necessary. I believe adding exceptions to the configuration applies them to the whole project.
I think it would be better if there is a single exception to disable it inline or within the file that occurs. 
https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments

packages/app-shell/src/components/Modal/index.jsx
```diff
@@ -88,10 +94,14 @@ const Modal = ({ modal, openModal }) => {
       openModal(MODALS.trialExpired)
     }
 
+    //Check if Pendo loads on the page - we don't want to show the OB Migration modal if there is a Pendo guide already visible
+    const isPendoModalVisible = window.pendo && !window.pendo.isGuideShown() || !window.pendo ? false : true;
```
dinostheo, line 98:
```suggestion
    const isPendoModalVisible = window.pendo && !window.pendo.isGuideShown() ? false : true;
```
I believe that we don't need the extra OR condition. Could it be that I'm missing something?

dinostheo (changes requested):
That's a great find @gisete. Using the pendo API to detect if it's visible seems like a solid approach to me 🌮 . 

I left some comments related to the eslint config and the `isPendoModalVisible`. Have a look at those before merging.

packages/app-shell/src/components/Modal/index.jsx
anafilipadealmeida: Is there any reason to go for the negation? Aren't all the cases included if we just use this:
`const isPendoModalVisible = window.pendo && window.pendo.isGuideShown();`

anafilipadealmeida: @gisete, do you need to cover any extra scenario or could `const isPendoModalVisible= window.pendo && window.pendo.isGuideShown()` be enough here?

anafilipadealmeida: I had deleted the comment by mistake, re-added it here ☝️ 😂   When creating boolean variables, you often don't need ternary conditions returning true and false, I usually either change the negative conditions to positive or use the double negation (eg. !!window.pendo instead of !window.pendo ? false : true )

.eslintrc.js
gisete: Hey @dinostheo, great question and I should have clarified this but I was getting so many formatting errors and I wasn't sure if we wanted to adopt all these rules so I disabled the ones that I was unsure of. I figured since we just started making lint a requirement we could maybe wait until Federico is back to discuss as a team what rules we want to enforce 🙂

packages/app-shell/src/components/Modal/index.jsx
gisete: Hey @anafilipadealmeida and @dinostheo, I admit it's a bit confusing 😄   My thinking here was that `!window.pendo` covers when we're working locally and Pendo doesn't load and `window.pendo && !window.pendo.isGuideShown()` check if Pendo loads and a Pendo modal is up . But feel free to suggest another approach if it makes more sense.

packages/app-shell/src/components/Modal/index.jsx
```diff
@@ -76,6 +77,11 @@ const ModalContent = ({ modal, closeAction }) => {
   }
 };
 
+ModalContent.propTypes = {
```
gmzjuliana, line 80: not a blocker but since we're moving towards typescript, it might be a good idea to start using it instead of propTypes going forward 🤔 maybe a good discussion for the working group @anafilipadealmeida

packages/app-shell/src/components/Modal/index.jsx
```diff
@@ -88,10 +94,14 @@ const Modal = ({ modal, openModal }) => {
...
 
+    //Check if Pendo loads on the page - we don't want to show the OB Migration modal if there is a Pendo guide already visible
+    const isPendoModalVisible = window.pendo && !window.pendo.isGuideShown() || !window.pendo ? false : true;
+
     //Migrate to OB modal
     const canMigrateToOneBuffer = user?.currentOrganization?.canMigrateToOneBuffer?.canMigrate;
     const hasDismissedMigrationModal = getCookie({ key: 'migrationModalDismissed' })
-    if (!hasDismissedMigrationModal && canMigrateToOneBuffer) {
+
+    if (!hasDismissedMigrationModal && canMigrateToOneBuffer && !isPendoModalVisible) {
```
gmzjuliana, line 104: not a blocker 😄 for readable you could consider storing these in a variable with a name that lets future devs know what all of these conditions boil down to. In this case could be something like:

```
let canMigrate = !hasDismissedMigrationModal && canMigrateToOneBuffer && !isPendoModalVisible
if (canMigrate) {...
```

gmzjuliana (approved):
LGTM 👍

packages/app-shell/src/components/Modal/index.jsx
anafilipadealmeida: If all you need to know is if the Pendo modal is up, we know that the modal will only be shown if pendo exists and isGuideShown is true. Then you can confidently say that in any other scenario the modal is not there (so no need to think about local vs production, as you are just focusing on pendo being defined and modal being up). By using this: `const isPendoModalVisible= window.pendo && window.pendo.isGuideShown()` in theory you are covering all the scenarios. Is this making any sense @gisete ?

gisete: You're totally right @anafilipadealmeida ! I think I got confused with another method I was using before to check if Pendo was available. I'll update it. Thanks for the feedback 🙂

packages/app-shell/src/components/Modal/index.jsx
gisete: Hey @gmzjuliana , good one to keep in mind. I noticed we actually don't use propTypes as default but I only used it here because ESlint was complaining about it. I'll discuss it with Federico for sure when he's back.

PR #103, opened 2021-08-23
Trial Expired Modal copy adjustment

gisete:
Update copy on the Trial Expired Modal. This includes update the list of features to show limitations of the Free plan instead.

image


dinostheo (approved):
@gisete this LGTM 🙌 . 
I see that linting is raising some errors during the CI. Could you have a look and fix those before merging?

PR #105, opened 2021-08-25
[FIX] Add eslitn cypress package

daisymarie128:
This removes many eslint errors when working in testing files

This PR installed a new package at the root level `"eslint-plugin-cypress": "^2.11.3",` and updates the `.eslintconfig` to handle test packages.
This was installed this to get rid of a large amount of eslint errors which are in the package due to test files using jest and cypress

This helps reduce/fix errors that appear if you run `yarn run lint` in the repo, from 674 problems down to 280 problems.

These are the types of eslint errors this fixes:
Screen Shot 2021-08-25 at 2 13 14 pm

Before adding this change working in test files looked like this:
Screen Shot 2021-08-25 at 1 58 23 pm

Before change this is how many problems `yarn lint` showed:
Screen Shot 2021-08-25 at 2 12 40 pm

After change this is how many problems `yarn lint` showed:
Screen Shot 2021-08-25 at 2 13 53 pm

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,8 +2,15 @@ module.exports = {
   env: {
     browser: true,
     es2021: true,
+    jest: true,
   },
-  extends: ['plugin:react/recommended', 'airbnb', 'prettier', 'prettier/react'],
+  extends: [
+    'plugin:react/recommended',
+    'airbnb',
+    'prettier',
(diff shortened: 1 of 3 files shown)


federicoweber (approved):
Thanks @daisymarie128 🙌

PR #108, opened 2021-08-26
Fix/cypress failing tests

daisymarie128:
This PR aims to fix the failing Cypress tests in our pipeline.

The tests failing in our PRs were also failing locally.

I think the root problem may have been that the intercepted call to `https://graph.buffer.com/` was pointing to the wrong url.
Locally by specifying the request should point to `https://graph.local.buffer.com/` seems to have fixed the issue.
**Pending** waiting to see if the tests will also pass in this PR

Here are screenshots of the tests passing locally:
Screen Shot 2021-08-26 at 2 45 39 pm


cypress.json
```diff
@@ -1,5 +1,10 @@
 {
   "video": false,
   "viewportWidth": 1440,
-  "viewportHeight": 900
+  "viewportHeight": 900,
+  "env": {
+      "CYPRESS_BASE_URL": "http://0.0.0.0:3000",
+      "REACT_APP_API_GATEWAY_URL": "https://graph.local.buffer.com/"
```
federicoweber, line 7: This are not needed. The workflow will automatically get all the variables defined in the workflow config as long as they are prefixed with `CYPRESS` 😉

.github/workflows/cypress.yml
```diff
@@ -13,7 +13,8 @@ jobs:
         run: yarn install && yarn add -W serve
       - uses: cypress-io/github-action@v2
         env:
-          CYPRESS_BASE_URL: http://0.0.0.0:3000 
+          CYPRESS_BASE_URL: http://0.0.0.0:3000
+          REACT_APP_API_GATEWAY_URL: https://graph.local.buffer.com/
```
federicoweber, line 17: We will need to prefix this with `CYPRESS` ([ref.](https://github.com/cypress-io/github-action#env)) if we want Cypress worker to pick it up.
But I don't think we need this in here, as it's injected in the app during the [build ](https://github.com/bufferapp/app-shell/blob/main/packages/app/package.json#L20)😉.

federicoweber (commented):
Hey @daisymarie128 looling forward to catch up on this. I have a feeling the GATEWAY URL this might be a red herring

cypress/fixtures/accountObFree.json
```diff
@@ -15,6 +15,11 @@
             "name":"test",
             "canEdit":true,
             "role":"admin",
+            "canMigrateToOneBuffer": {
```
federicoweber, line 18: Ah this one make sense, if the schema is different GraphQL won't match.
It does explain why it started failing once we merged https://github.com/bufferapp/app-shell/pull/77

federicoweber:
Great work fixing this. Thanks @daisymarie128 🙌

federicoweber (approved):
@daisymarie128 let's :shipit: 🙌

PR #109, opened 2021-09-01
fix isMultiProductCustomer trait

federicoweber:
Follow up to https://github.com/bufferapp/core-authentication-service/pull/1256/, to make sure we are not tracking as MP an org that is trialing.


dinostheo:
@federicoweber code wise this LGTM. I left a comment in the server side PR https://github.com/bufferapp/core-authentication-service/pull/1256#discussion_r700917804, regarding trials if we need to exclude them or include them. 
Based to your answer, we might have to modify this PR or the server side PR.

hamstu (approved):
This looks good Federico, LGTM! 👍🏻

PR #110, opened 2021-09-07
Track when the OB migration modal is viewed

gisete:
Track views on the various stages of the OB migration modal. 

NOTE: I've added the tracker to the first 3 stages but left out of the last (Success) one as I thought it might not be valuable to track views for it. Let me know if that's not the case 🙂

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/PaidMigration/EssentialsPlan/index.jsx
+++ b/packages/app-shell/src/components/Modal/modals/PaidMigration/EssentialsPlan/index.jsx
@@ -24,13 +24,13 @@ export const EssentialsPlan = ({ features }) => {
   const { cta, ctaButton } = data || {};
 
   useEffect(() => {
-    const productFromPath = getActiveProductFromPath();
+    const product = getActiveProductFromPath();
 
     useTrackPageViewed({
       payload: {
         name: 'Migrate to OB Modal',
         title: 'Value',
-        product: productFromPath,
+        product,
         cta,
         ctaButton,
       },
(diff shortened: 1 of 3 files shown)


packages/app-shell/src/components/Modal/modals/PaidMigration/EssentialsPlan/index.jsx
```diff
@@ -1,19 +1,38 @@
...
+import {
+  useTrackPageViewed,
+} from '../../../../../common/hooks/useSegmentTracking';
 
 import * as styles from './style';
 
 export const EssentialsPlan = ({ features }) => {
+  const currentUser = useContext(UserContext);
+  const { data } = useContext(ModalContext);
+  const { cta, ctaButton } = data || {};
```
federicoweber, line 22: I don't think we are passing any CTAs yet, but good call on preparing for those, they could be useful 👍.

federicoweber (commented):
Hey @gisete, thanks for working on this.
The logic looks good to me, the only thing that might be worth adding is the product attribute; you can get that using `getActiveProductFromPath`.

And you can find a list of all supported attributes in [here](https://github.com/bufferapp/tracking-plan/blob/master/tracking-plans/buffer-tracking-plan/events/product/shared/page-viewed.yaml), if you want to add some extra information 😉 .

For the title, name combinations I would suggest you to reach out to @tomredman, as he will be the one querying this data the most 🙂.

gisete:
Hey @federicoweber , thanks for the tip about adding the `product` attribute. I did notice though that `product` was already being passed to the segment event. I went back to the code and noticed this: https://github.com/bufferapp/app-shell/blob/main/packages/app-shell/src/common/hooks/useSegmentTracking.js#L52-L57

Does this mean we're adding `product` by default and maybe don't need to add to the event here again?

federicoweber:
> Hey @federicoweber , thanks for the tip about adding the `product` attribute. I did notice though that `product` was already being passed to the segment event. I went back to the code and noticed this: https://github.com/bufferapp/app-shell/blob/main/packages/app-shell/src/common/hooks/useSegmentTracking.js#L52-L57
> 
> Does this mean we're adding `product` by default and maybe don't need to add to the event here again?

Ah sorry @gisete, I forgot we were injecting that for all the events 😅. That's all good then 🙂

PR #111, opened 2021-09-09
[SP-144] add start page to navigator

gmzjuliana:
This PR checks if a user is admin and if they have the start page feature flip enabled to show them Start Page in the Navigator


packages/app-shell/src/components/NavBar/index.jsx
```diff
@@ -250,12 +250,14 @@ const NavBar = React.memo((props) => {
   const shouldDisplayInviteCTA =
     user?.currentOrganization?.shouldDisplayInviteCTA;
   if (user.currentOrganization.billing) {
-    // eslint-disable-next-line no-unused-vars
```
gmzjuliana, line 253: i thiiiink we don't need this anymore :)

federicoweber: looks like we do, `subscription` is breaking in the linter.
Can we remove `subscription`?

packages/app-shell/src/components/NavBar/components/NavBarProducts/NavBarProducts.jsx
```diff
@@ -71,50 +71,63 @@ const NewLabel = styled.span`
...
-    label: 'Analytics',
-    isNew: false,
-  },
-  {
-    id: 'engage',
-    label: 'Engagement',
-    isNew: false,
-  },
-];
+const NavBarProducts = ({ activeProduct, showStartPage }) => {
```
federicoweber, line 74: let's compute `showStartPage` in here, instead of passing it down as a prop.
It will make it easier to understand what is going on.

federicoweber (changes requested):
Hey @gmzjuliana thanks for working on this one.
Left a few small comments I would like to address before merging this in, but otherwise, it's looking good to me.

PR #112, opened 2021-09-13
Add alias to 'common' directory

gisete:
This PR modifies the webpack config file to add an alias for the `common` directory so we don't have to use relative paths. [[documentation]](https://webpack.js.org/configuration/resolve/#resolvealias). 

Also added a new `jsconfig.json` file to the project so VS Code's auto complete can work with webpack aliases. [Source](https://code.visualstudio.com/docs/languages/jsconfig#_using-webpack-aliases).
Example:
Screen Recording 2022-04-14 at 04 06 30 PM

_note: modified the original approach in this PR so we don't need to keep updating aliases as we add more subfolders to `common`_


gisete:
Hey @federicoweber, seeing how long the relative imports were was kind of bugging me 😄  so I came up with this solution. 

For the VSCode solution, I know the team might be using different editors so not sure if this will have an impact on how y'all work.

federicoweber (approved):
Ah great call @gisete definitely makes it easier to cleaner 💯

cheahenry (approved):
Nice, this looks way cleaner. Thanks for tackling this!

daisymarie128 (approved):
I'm in love with this PR 😍  @gisete big thank you!! 🌮

PR #116, opened 2021-09-22
add sc_attr to webpack, add namespace to babel config and move styled…

gmzjuliana:
Adding namespacing in babel plugins and sc_attr to the build to avoid style collisions

Heads up, I've also moved styled-components from devDependencies to dependencies


gmzjuliana:
closed for #118

PR #117, opened 2021-09-22
improve animation of the plan selector confirmation modal [PUB-3836]

federicoweber:
Prevent double rendering of the confirmation modal, that was caused by Account query refreshing after the mutation was completed.
fixes: [PUB-3836](https://buffer.atlassian.net/browse/PUB-3836)

### DIFF
--- a/packages/app-shell/src/common/hooks/useUpdateSubscriptionPlan.js
+++ b/packages/app-shell/src/common/hooks/useUpdateSubscriptionPlan.js
@@ -17,6 +17,7 @@ const useUpdateSubscriptionPlan = ({
     UPDATE_SUBSCRIPTION_PLAN,
     {
       refetchQueries: [{ query: QUERY_ACCOUNT }],
+      awaitRefetchQueries: true,
     }
   );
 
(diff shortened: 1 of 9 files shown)


packages/app-shell/src/common/hooks/useUpdateSubscriptionPlan.js
```diff
@@ -17,6 +17,7 @@ const useUpdateSubscriptionPlan = ({
     UPDATE_SUBSCRIPTION_PLAN,
     {
       refetchQueries: [{ query: QUERY_ACCOUNT }],
+      awaitRefetchQueries: true,
```
federicoweber, line 20: Note: `awaitRefetchQueries` will force the mutation to return only once the query refetch is completed.
But I noticed that the execution is not really synchronous, and in some cases that account updates slightly faster than the mutation returns, in those cases, we still see a small glitch during the transition, caused by a rerender.

I think this is an acceptable compromise for the time being, as at least it's still much better that before.
Also, we shouldn't rely on `refetchQueries` to begin with, and instead return the data in the mutation results, that should completely get rid of any issue caused by this racing condition 😉.

PR #122, opened 2021-10-07
[FIX] Change copy for Migrate to Essentials

daisymarie128:
This change removes confusion around using the word monthly when the
user is on a yearly plan

TIcket: https://buffer.atlassian.net/browse/GROW-261

Manual testing:
The below scenarios were both tested by impersonating 2 different users locally in app shell against production.

- [x]  A user who can migrate to the essentials plan and is currently on a yearly subscription:

Screen Shot 2021-10-11 at 9 42 55 am

- [x]  A user who can migrate to the essentials plan and is on a monthly subscription:

Screen Shot 2021-10-11 at 10 00 34 am

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/PaidMigration/EssentialsPricing/index.jsx
+++ b/packages/app-shell/src/components/Modal/modals/PaidMigration/EssentialsPricing/index.jsx
@@ -155,13 +155,13 @@ export const Content = ({
             <TotalPrice>
               <sup>$</sup>
               <Text type="h2" as="p">
-                {migrationPreview.preview.currentPlan.interval === 'month'
+                {migrationPreview.currentPlan.interval === 'month'
                   ? migrationPreview.migrationSummary.totalPrice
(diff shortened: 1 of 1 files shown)


packages/app-shell/src/components/Modal/modals/PaidMigration/EssentialsPricing/index.jsx
```diff
@@ -151,19 +155,23 @@ export const Content = ({
             <TotalPrice>
               <sup>$</sup>
               <Text type="h2" as="p">
-                {migrationPreview.migrationSummary.totalPrice}
+                {migrationPreview.currentPlan.interval === 'month'
+                  ? migrationPreview.migrationSummary.totalPrice
```
dinostheo, line 159: Just to mention that there is a bug that we need to fix, as for Free plan `migrationPreview.imgrationSummary` is `undefined`. However, this is not part of this PR and it is currently as such in production. 

cc @cheahenry

PR #125, opened 2021-10-20
append query parameters to logout redirect

federicoweber:
This ensures that any query parameters are properly passed down to the redirect URL for users without an active session.

You can test this out in here https://fix-logout-redirect-appshell.dev.buffer.com/?foo=foo&bar=bar

## Notes: 
- This is a follow-up to the work started in https://github.com/bufferapp/buffer-publish/pull/1761.
- This issue was reported by @mallen5311 via Slack: https://buffer.slack.com/archives/CHRPK6R6D/p1617728115071500
- This is passing down the parameters to the login service, but those are getting stripped down when logging in, so we will need additional work if we want to persist those parameters after the login.

### DIFF
--- a/packages/app-shell/src/components/NavBar/index.jsx
+++ b/packages/app-shell/src/components/NavBar/index.jsx
@@ -47,6 +47,10 @@ export function getProductPath(baseUrl) {
   return productPath;
 }
 
+export function getQueryParameters(baseUrl,) {
+  return baseUrl.match(/\?(?<query>.*)$/)?.groups?.query
+}
+
 function getRedirectUrl(baseUrl) {
   const productPath = getProductPath(baseUrl);
   return `https://${productPath}.buffer.com`;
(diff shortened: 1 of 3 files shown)


packages/app-shell/src/components/NavBar/index.jsx
```diff
@@ -56,7 +60,14 @@ export function getLogoutUrl(baseUrl = '') {
   const productPath = getProductPath(baseUrl);
   return `https://login${
     productPath.includes('local') ? '.local' : ''
-  }.buffer.com/logout?redirect=${getRedirectUrl(baseUrl)}`;
+  }.buffer.com/logout?redirect=${getRedirectUrl(baseUrl)}&${getQueryParameters(baseUrl)}`;
+}
+
+export function getLoginUrl(baseUrl = '') {
```
federicoweber, line 66: I've added a `/login` redirect as that looks like a better choice when redirecting logged-out users 😉.

packages/app-shell/src/components/NavBar/index.jsx
```diff
@@ -47,6 +47,11 @@ export function getProductPath(baseUrl) {
   return productPath;
 }
```
msanroman, line 49: I would move all the URL methods to its own module, and out of the React component file. Would reflect even better since the `NavBar` tests are now testing methods that have nothing to do with the NavBar (nor should they).

packages/app-shell/src/exports/Navigator/index.jsx
```diff
@@ -63,7 +63,7 @@ export const Navigator = React.memo(({ apolloClient, channels }) => {
   if (
     networkErrors?.some((err) => err.extensions?.code === 'UNAUTHENTICATED')
   ) {
-    window.location.assign(getLogoutUrl(window.location.href));
+    window.location.assign(getLoginUrl(window.location.href));
```
msanroman, line 66: With those URL methods somewhere else other than React components, I might even create a `redirect` sort of method that does the `window` interactions under the hood. That would make all React components a bit "dumber" and isolated from all those details.

msanroman (approved):
Great work, Fed! I've added some comments because I think it's time we move some of these methods elsewhere, but the overall implementation feels great to me!

PR #134, opened 2021-12-02
Fix/cleanup unused interval base price

dinostheo:
Cleaning 2 properties that are exclusive to the Plan Selector and we. don't use them anymore. After cleaning it here, I will continue with the API and the Plan Options in the server side. 

Plan Options is an exclusive object for the Plan Selector and if there are properties that we don't use, because the UI changed, it would be good to clean them up 😃.


cheahenry (approved):
Thanks for starting the cleanup!

PR #135, opened 2021-12-09
[GROW-306] extract trial banner copy to a function

cheahenry:
Extracted the copy to a function and added a test.

The `planName` is coming from here: 
https://github.com/bufferapp/core-authentication-service/blob/bbe172d413aa32ac744effc155665e3b0f9bd332/shared/services/billing/interfaces.ts#L23-L28

Tested locally:
Screen Shot 2021-12-09 at 12 11 20 PM


dinostheo (approved):
LGTM @cheahenry 🚢

PR #136, opened 2021-12-09
[GROW-306] Clean up summary modal code

cheahenry:
Clean up summary modal code.

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/Summary/index.jsx
+++ b/packages/app-shell/src/components/Modal/modals/Summary/index.jsx
@@ -1,9 +1,6 @@
 import React from 'react';
 import Text from '@bufferapp/ui/Text';
 import Coupon from '@bufferapp/ui/Icon/Icons/Coupon';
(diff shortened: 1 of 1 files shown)


dinostheo:
@cheahenry that's a good change, but I believe we don't use this anymore. I believe we never cleaned this logic. 

To give you some context, in the early days of the Plan Selector when we were downgrading we were displaying a warning banner to the user to let them know what they will be missing moving to a lower plan. 
However, we deprecated that warning on the server side, some time ago. I might be wrong, but I believe this functionality was there to find out if we need to display that warning or not.

cheahenry:
@dinostheo yeah I just realized and just pushed some clean up :D

dinostheo (approved):
Thanks for the cleanup 🙌 
We should make it a habit and clean code that is not needed quickly, otherwise we forget 😄

cheahenry:
I messed up my branch, gonna close this PR and open a new one 😅

PR #138, opened 2021-12-13
Task/grow 310 agency plan selector

daisymarie128:
### Purpose
This PR aims to add the new Agency plan to the Plan Selector Modal.

Jira ticket: https://buffer.atlassian.net/browse/GROW-310

### Changes
This PR introduces changes to the logic and view of the Plan Selector.
- The Free Plan card has been removed and appears as a small footer section. Which required removing some logic from plan filtering. This  section only shows if:
   - The user is not on a Free Plan and has the ability to downgrade to Free
   - Will not show for Free users
- Agency Plan card has been added to the modal
- New hook `useDefaultSelectedPlan` created to simplify logic removing Free plan conditions
- Minor refactoring to styling for the PlanSelector modal

**User already on the Essentials Plan w/ monthly selected:**
Screen Shot 2021-12-14 at 12 03 37 pm

**User already on the Essentials Plan w/ yearly selected:**
Screen Shot 2021-12-14 at 12 03 33 pm

**User already on the Free Plan w/ monthly selected:**
Screen Shot 2021-12-13 at 12 36 31 pm

**User already on the Free Plan w/ yearly selected:**
Screen Shot 2021-12-14 at 11 54 44 am

**User already on the Essentials Plan w/ Current plan selected:**
Screen Shot 2021-12-14 at 11 55 11 am

**User without agencyPlan in Feature Flips**
Screen Shot 2021-12-16 at 2 12 43 pm

### TODO

- [x]  Write new tests for new logic
- [x] Manually test UI is working as expected
- [x] Manually test Upgrading and Changing to Agency plan (I'm unaware how to do this locally atm, will bring this up in call 
with Dinos 😄  )

### TESTING SCENARIOS
**With agencyPlan FF ON**
- [x] User on a Free plan sees Essentials, Essentials +team, Agency.
- [x] Users on Free plan do NOT see "Downgrade to Free" banner in footer of modal
- [x] Users on Essentials plan see "Downgrade to Free" banner in footer of modal
- [x] User already on the Agency sees Agency as default selected in modal
- [x] User on the Free plan sees Agency plan
- [x] User on the Free plan has Essentials plan selected by default when opening the modal

**With agencyPlan FF OFF**
- [x] Users see Free, Essentials and Essentials +team in the modal.
- [x] User do not see the Agency plan
- [x] User on the Free plan has Essentials plan selected by default when opening the modal
- [x] Users on the Free plan open the modal and have the free plan selected by default


dinostheo (approved):
This looks great to me @daisymarie128 
I skimmed through the CSS part as I don't have enough knowledge to judge it 😅 
I do have some remarks regarding the Free plan. I'm wondering if this is the way we go design wise that we should stop sending the Free plan option from the server side, instead of trying to filter it. 
This doesn't block this PR at the moment, but if we remove the Free option from the server side then we should clean up the code that filters it out as well.

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useFeatureFlip.js
```diff
@@ -0,0 +1,7 @@
+const useFeatureFlip = (user, featureFlip) => {
+  if (!user || !user.featureFlips) return false;
+
+  return user.featureFlips.includes(featureFlip);
+};
```
hamstu, line 5: This is a great extraction of logic! My only comment is that naming it like a hook (`use***`) isn't necessary since it doesn't use any hooks itself (useState, useRef, etc.) so this could just be named and used as a plain function.

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useFilteredListOfPlans.js
```diff
@@ -0,0 +1,9 @@
+const useFilteredListOfPlans = (planOptions, planToExclude) => {
```
hamstu, line 1: Same as prev. – this can be a plain function :)

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useFilteredListOfPlans.test.jsx
```diff
@@ -0,0 +1,45 @@
...
+];
+
+const listOfFilteredPlanOptions = [
+  { planId: 'essentials', planInterval: 'month', isCurrentPlan: false },
+  { planId: 'team', planInterval: 'year', isCurrentPlan: true },
+];
+
+describe('useFilteredListOfPlans', () => {
+  it('should remove all plans which match the passed in value to filter', () => {
+    const { result } = renderHook(() =>
```
hamstu, line 17: This test will be easier since you won't have to render a hook, but just test the function.

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useDefaultSelectedPlan.js
```diff
@@ -0,0 +1,18 @@
+import useFilteredListOfPlans from './useFilteredListOfPlans';
+import useFeatureFlip from './useFeatureFlip';
+
+const useDefaultSelectedPlan = (planOptions, user) => {
```
hamstu, line 4: Coming back to this after looking at the two methods it calls, and I don't see any base hooks in use. Can be a plain function instead.

hamstu (changes requested):
This looks great Daisy, I love how things are broken down and well tested. The only thing I noticed, as you'll see in my comments is a lot of your hooks can actually just be regular functions. The only reason to create a custom hook would be to encapsulate logic of another hook (often one of the 'base' hooks that React ships with.) Let me know if that makes sense!

hamstu (approved):
✅   Awesome stuff Daisy!

PR #140, opened 2022-01-11
Task/grow 310 agency plan refactored

daisymarie128:
### Purpose
This PR aims to re add the agency plan selector work to production.
This PR was originally merged on the 07/01/2022: https://github.com/bufferapp/app-shell/pull/138

This PR was then reverted: https://github.com/bufferapp/app-shell/pull/139

Link to first mention of customers scenario breaking: https://buffer.slack.com/archives/C0KS2S8D7/p1641549127005700
Link to PagerDuty/slack thread discussing the issue and the revert: https://buffer.slack.com/archives/C0KS2S8D7/p1641555856005900

I've made a short loom video explaining the new code I've added to fix this issue. : https://www.loom.com/share/36672ec9813e4dd083ebb810c1a2ac1b?sharedAppSource=personal_library

This PR also contains new tests which cover the use case mentioned in the threads above.
Screen Shot 2022-01-11 at 2 55 25 pm

Let me know if you have any questions 🙂

Everything below this line is what was in the initial PR description before the PR had to be reverted:
------

This PR aims to add the new Agency plan to the Plan Selector Modal.

Jira ticket: https://buffer.atlassian.net/browse/GROW-310

### Changes
This PR introduces changes to the logic and view of the Plan Selector.
- The Free Plan card has been removed and appears as a small footer section. Which required removing some logic from plan filtering. This  section only shows if:
   - The user is not on a Free Plan and has the ability to downgrade to Free
   - Will not show for Free users
- Agency Plan card has been added to the modal
- New hook `useDefaultSelectedPlan` created to simplify logic removing Free plan conditions
- Minor refactoring to styling for the PlanSelector modal

**User already on the Essentials Plan w/ monthly selected:**
Screen Shot 2021-12-14 at 12 03 37 pm

**User already on the Essentials Plan w/ yearly selected:**
Screen Shot 2021-12-14 at 12 03 33 pm

**User already on the Free Plan w/ monthly selected:**
Screen Shot 2021-12-13 at 12 36 31 pm

**User already on the Free Plan w/ yearly selected:**
Screen Shot 2021-12-14 at 11 54 44 am

**User already on the Essentials Plan w/ Current plan selected:**
Screen Shot 2021-12-14 at 11 55 11 am

**User without agencyPlan in Feature Flips**
Screen Shot 2021-12-16 at 2 12 43 pm

### TODO

- [x]  Write new tests for new logic
- [x] Manually test UI is working as expected
- [x] Manually test Upgrading and Changing to Agency plan (I'm unaware how to do this locally atm, will bring this up in call 
with Dinos 😄  )

### TESTING SCENARIOS
**With agencyPlan FF ON**
- [x] User on a Free plan sees Essentials, Essentials +team, Agency.
- [x] Users on Free plan do NOT see "Downgrade to Free" banner in footer of modal
- [x] Users on Essentials plan see "Downgrade to Free" banner in footer of modal
- [x] User already on the Agency sees Agency as default selected in modal
- [x] User on the Free plan sees Agency plan
- [x] User on the Free plan has Essentials plan selected by default when opening the modal

**With agencyPlan FF OFF**
- [x] Users see Free, Essentials and Essentials +team in the modal.
- [x] User do not see the Agency plan
- [x] User on the Free plan has Essentials plan selected by default when opening the modal
- [x] Users on the Free plan open the modal and have the free plan selected by default


hamstu (approved):
This looks great @daisymarie128! 🚀 

Once merged I wonder if we can force a run of the synthetic test in Datadog — the same one that caught the issue last time (I'm not entirely sure if this is possible, but I think it would be, might be good to ask @federicoweber)

packages/app-shell/src/components/Modal/utils.js
```diff
@@ -28,3 +28,35 @@ export function shouldShowFreeUserStartTrialPrompt(user) {
...
+}
+
+export function userHasFeatureFlip(user, featureFlip) {
+  if (!user || !user.featureFlips) return false;
+
+  return user.featureFlips.includes(featureFlip);
+}
+
+export function getDefaultSelectedPlan(planOptions, user, isUpgradeIntent) {
+  const featureFilpAgencyPlan = userHasFeatureFlip(user, 'agencyPlan'); // TODO:REMOVE_WITH_FF:agencyPlan
```
dinostheo, line 47: Non blocking for this PR as it is used consistently 😄. However, there is a typo the variable here `featureFilpAgencyPlan` instead of `featureFlipAgencyPlan`

dinostheo (approved):
Thanks for fixing it and for the demo @daisymarie128. It looks great.

federicoweber:
> This looks great @daisymarie128! 🚀
> 
> Once merged I wonder if we can force a run of the synthetic test in Datadog — the same one that caught the issue last time (I'm not entirely sure if this is possible, but I think it would be, might be good to ask @federicoweber)

Hey @hamstu, as far as I know the synthetic tests are still running on DD, every 30 minutes, and it's also possible to manually run those. But mind if I ask what is the reason you want to do that?

hamstu:
Thanks @federicoweber! Oh if its every 30 then we can probably just wait, but I mention it since we had an issue with a previous PR that the synthetic caught, but I don't think we noticed it right away so it wasn't fixed until a revert some time later. 👍🏻

federicoweber:
> Thanks @federicoweber! Oh if its every 30 then we can probably just wait, but I mention it since we had an issue with a previous PR that the synthetic caught, but I don't think we noticed it right away so it wasn't fixed until a revert some time later. 👍🏻

Got it. Thing is, we shouldn't rely on those tests to validate a change, they are really meant as a last resort because they are running against the production, so if anything is off there it's also off for our users.
I would suggest replicating the breaking condition with a cypress test instead so that we could spot the issue before merging 😉.

daisymarie128:
> @daisymarie128 this should say "Includes 10 social channels" Screenshot 2022-01-11 at 16 03 57@2x

Good catch @tomredman I've updated this copy now see screenshots below

@dinostheo Wanted to get your opinion on this as well. Is this something we could amend on the BE? I took a quick look at the code but wasn't too sure if `channelsQuantity` actually comes from how many channels a user currently has connected 🤔 

Screen Shot 2022-01-13 at 8 22 03 am
Screen Shot 2022-01-13 at 8 22 09 am

dinostheo:
@daisymarie128 `channelQuantity` is the channels connected. However, we could return the value from the server side for Agency to show `10` if the connected channels are <= than 10 and then actual connected channels, if the value is above it. 
I can make this fix tomorrow morning.

dinostheo:
@daisymarie128 if this is what blocks this PR, please merge it, since Agency is not live for users except some test ones and I can correct the server side on time.

dinostheo:
@daisymarie128 regarding the `channelsQuantity`, I deployed a fix on the server side to provide the correct number based on the plan and the channels connected for an organization. Therefore I pushed a commit here that removes the ternary that checks what the plan is and tries to default to `10` channelsQuantity for Agency plans. 

The expected result should be the following.

**All plans (excluding Agency)**
 * With `0` channels connected: `channelsQuantity=1` 
 * With `1` channel connected: `channelsQuantity=1`
 * With `15` channels connected: `channelsQuantity=15`
 
**For the Agency plan**
 * With `0` channels connected: `channelsQuantity=10` 
 * With `1` channel connected: `channelsQuantity=10`
 * With `15` channels connected: `channelsQuantity=15`
 
Screen Shot 2022-01-13 at 13 08 37
 The above screenshot shows how the Plan Selector will look like with the Agency plan option selected and 12 channels connected.

daisymarie128:
@dinostheo Thanks for updated that on the BE and removing the conditional. I tested it locally as well and it all works. I'm going to merge this now 🎉 🌮

PR #142, opened 2022-01-18
Fix/grow 328 downgrade cta

daisymarie128:
This PR fixes the `Downgrade to our free plan` button in the new plan selector.
Previously this button/text used a placeholder link which did nothing. After this change, the user can now Downgrade themselves to the free plan.

🛑  While doing manual testing for this I discovered that this scenario was also *broken. *Broken in this case meaning, the whole app-shell/navigator would stop working, throw and error and disappear until the page was refreshed.
(I will link to a video of some of the debugging done during this)

🏴  I'm not actually sure how long this has been broken. While trying to fix it, I couldn't see any connection to the new Agency plan code affecting this, so am unaware if this was caused by something older 🤔  Interested to hear any other thoughts as to why this was broken 😅

Connected to this Jira ticket: https://buffer.atlassian.net/browse/GROW-328

**Video of debugging start 14-day trial not working:** https://loom.com/share/efbc82b93f044f3fb51ab0503dcbf59f
**Videos of manually testing users with plan selector modal:**
Note I made these videos just to show how/what I've manually tested, I'm still learning new scenarios to always consider, and if I'm doing anything wrong/or which could be improved do let me know! 🤓 

**Agency plan to Free plan to Essentials plan**
https://www.loom.com/share/4aaca10aac9b4c1692cb31477c790fb6?sharedAppSource=personal_library

**Free plan upgrade to essentials plan**
https://www.loom.com/share/0c4578e03ca443a4988eaf6a25b051f9?sharedAppSource=personal_library

**User able to still start a 14-day free trial**
See the end of this video for the working test
https://loom.com/share/efbc82b93f044f3fb51ab0503dcbf59f

**What a user sees if they are on a free plan already**
Screen Shot 2022-01-18 at 4 57 16 pm

**What a user sees if they are on the team14 day trial of team plan**
Screen Shot 2022-01-18 at 4 54 29 pm

**What a user sees if they are on the agency plan already**
Screen Shot 2022-01-18 at 4 59 58 pm

**What a user sees if they have downgraded to the free plan halfway through billing cycle**
Screen Shot 2022-01-18 at 4 55 42 pm


packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
```diff
@@ -183,7 +185,21 @@ export const PlanSelectorContainer = ({
           updateSelectedPlan={updateSelectedPlan}
           monthlyBilling={monthlyBilling}
         />
-        {featureFilpAgencyPlan && !isFreePlan && <FreePlanSection />}
+        {featureFilpAgencyPlan && !isFreePlan && (
+          <FreePlanSection
+            ctaAction={() => {
```
cheahenry, line 190: Nitpick: I am not a fan of the naming `ctaAction` as its generic. Something including downgrade to free would be better for code readability.

cheahenry (approved):
LGTM, just a nitpick comment.
Thanks for working on this :)

dinostheo (approved):
Thanks so much for fixing the downgrade issue and the start trial as well 🙌 
About the latter, I followed your instructions on the recording and I was able to reproduce it for signups with and without the Agency experiment. 

My assumption here is that most probably users ran into this issue, but since it does start a proper trial and refreshing the page fixes it, we didn't get any reports.

PR #144, opened 2022-01-20
[GROW-177] Do not remove the modal hash from the URL and return it to login redirect

dinostheo:
We have some hashes that we can use in the buffer URLs which will trigger the Plan Selector and the Plan Migrator modal. Those are often used by customer advocates in order to provide a link to the users that they can use to self upgrade/downgrade or migrate. 
When a user is not authenticated we redirect them to the login service, which receives the current URL slightly modified (and verified) as a redirect URL. 
So far by trimming the hash from the URL, we were preventing the modal to open after the user have logged in successfully. 

This requires a small change in the account app as well since it overrides the redirect of the app shell. 

https://buffer.atlassian.net/browse/GROW-177

NOTE: Most of the changes are prettier formatting. The actual code changes are in packages/app-shell/src/common/hooks/useModal.js
and packages/app-shell/src/components/NavBar/index.jsx

### DIFF
--- a/packages/app-shell/src/components/NavBar/index.test.js
+++ b/packages/app-shell/src/components/NavBar/index.test.js
@@ -1,27 +1,67 @@
-import { getQueryParameters, getLoginUrl } from './index'
+import { getQueryParameters, getLoginUrl, getLogoutUrl } from './index';
 
 describe('NavBar', () => {
   describe('getQueryParameters', () => {
     it('return an empty strin if no query parameters', () => {
-      const params = getQueryParameters('http://buffer.com')
-      expect(params).toEqual('')
-    })
+      const params = getQueryParameters('http://buffer.com');
+      expect(params).toEqual('');
+    });
+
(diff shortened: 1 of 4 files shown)


packages/app-shell/src/components/NavBar/index.test.js
```diff
@@ -1,27 +1,67 @@
-import { getQueryParameters, getLoginUrl } from './index'
+import { getQueryParameters, getLoginUrl, getLogoutUrl } from './index';
 
-describe('NavBar', () => {
+describe.only('NavBar', () => {
```
cheahenry, line 3: Did you forget to remove this `only`?

dinostheo: Yes, I did 😅 . 
Thanks for catching this @cheahenry

cheahenry (approved):
LGTM :)

PR #145, opened 2022-01-20
Clean the account-app menu item url

dinostheo:
While testing the redirect changes I introduced in a previous PR (https://github.com/bufferapp/app-shell/pull/144), I noticed that the account menu item is passing a username and a redirect parameter that to my understanding does not seem useful. Additionally, it has the potential to create a bit weird recursive redirect URLs, like the following.
`https://account.buffer.com/?redirect=https://account.buffer.com/?redirect=https://account.buffer.com/?redirect=https://analyze.buffer.com&username=undefined`

The above is the result of consecutive clicks on the account menu item.


cheahenry (approved):
Good find!

PR #150, opened 2022-02-08
[UPDATE] Util funcs to calculate slots total pricing

daisymarie128:
Small util function we will use in the billing summary component. Includes unit tests


packages/app-shell/src/components/Modal/utils.test.jsx
```diff
@@ -290,23 +291,55 @@ describe('Modal - utils', () => {
...
+          ...mockUserData.currentOrganization,
+          channels: [],
+        },
+      });
       const result = shouldShowChannelConnectionPrompt(noChannelsUser);
       expect(result).toBeFalsy();
     });
   });
+
+  describe('calculateTotalSlotsPrice', () => {
```
daisymarie128, line 330: These are the only new tests. The above changes are just auto linting on the file.

hamstu (approved):

PR #152, opened 2022-02-14
[NEW-COMPONENT] ChannelsCounter

daisymarie128:
🚢 

This PR adds a new component available for the Summary component.
Required for the new SBB work.

This component is not actually implemented into where customers can see/use yet so is safe to merge
Screen Shot 2022-02-14 at 10 05 02 am
.


gisete (commented):
Way to ship a new component! 🚢  🚀

PR #157, opened 2022-02-16
Task/grow 344 setup sbb components

daisymarie128:
🚢 

This PR creates a new render function for the SBB feature flip. It will display the new Summary section which allows users to pre-select the number of channels they want.

Currently, this just creates the render functions but doesn't call it yet (Haven't made it call it yet because all the functionality is not hooked up yet) will do it in later PR

Screen Shot 2022-02-16 at 2 13 37 pm
Screen Shot 2022-02-16 at 2 13 27 pm
Screen Shot 2022-02-16 at 2 13 17 pm


packages/app-shell/src/components/Modal/utils.js
```diff
@@ -73,6 +73,12 @@ export function getDefaultSelectedPlan(planOptions, user, isUpgradeIntent) {
   return defaultSelectedPlan;
 }
 
-export function calculateTotalSlotsPrice(numberOfSlots, slotPrice) {
+export function calculateTotalSlotsPrice(numberOfSlots, slotPrice, planId) {
+  if (planId === 'agency') {
+    if (numberOfSlots <= 10) {
+      return slotPrice;
+    }
+    return numberOfSlots * (slotPrice / 12);
```
daisymarie128, line 81: @dinostheo @cheahenry wdyt about in the future a more relatable price coming back from the BE for agency? 
Currently, the basePrice for agency is 100/120 and this calculation doesn't really work for that and it's a bit tediuos adding in the conditional to check if its agency and hard code in $5/6 in particular places :)

dinostheo: @daisymarie128 @gisete @cheahenry when the Agency plan is selected the default quantity should be 10 for SBB. We'll take of that on the server-side, when a user upgrades to Agency with less than 10 channels, or when they subscribe to Agency directly that they get 10 channel slots. 

However, on the client-side, the channel slot modal should default (set a minimum limit) as well. It should not allow the users to go under the 10 channel slots. 
So the calculation should always be for Agency:
`flatFee + (excesSlots * montlySlotPrice)` and for a yearly cycle is the same formula by `*12`.

Perhaps, we could generalize this by sending the price values for all plans:
e.g. 

For `Essentials`:
```
monthlyPlanFlatFee = 0
channelQuantity = 9
chargableQuantity = 9
monthlyPricePerQuantity = 5
```

Using the above price calculation `monthlyPlanFlatFee + (chargableQuantity * monthlyPricePerQuantity)` the total monthly cost would be `45` and the yearly `45 * 12 = 540`

For `Agency` 1st variant `quantity < 10`:
```
monthlyPlanFlatFee = 100
channelQuantity = 9
chargableQuantity = 0
monthlyPricePerQuantity = 5
```

Using the above price calculation `monthlyPlanFlatFee + (chargableQuantity * monthlyPricePerQuantity)` the total monthly cost would be `100` and the yearly `100 * 12 = 1200`

For `Agency` 2nd variant `quantity > 10`:
```
monthlyPlanFlatFee = 100
channelQuantity = 15
chargableQuantity = 5
monthlyPricePerQuantity = 5
```

Using the above price calculation `monthlyPlanFlatFee + (chargableQuantity * monthlyPricePerQuantity)` the total monthly cost would be `100 + 25 = 125` and the yearly `125 * 12 = 1500`

In that way, I believe we normalize on the server-side all the pricing and the client acts nearly agnostic to the plan. We could also send from the server-side a value, perhaps called `minimumSlotQuantity`, which would be `10`  for the Agency plan and `1` for all the others.

Let me know what you think?

PR #169, opened 2022-02-24
Task/grow 389 fix upgrade bug

daisymarie128:
👀  Could someone do a quick check over this PR as it is to fix a bug which was found in production 😃 

This PR applies a fix to a recent incident that occurred on production. 

This affected users clicking on the Upgrade CTA in the Navigator and would cause the navigator to disappear and break completely.

This fix also contains a new Cypress test to catch this user flow in the future.

Slack thread with more conversation surrounding this: https://buffer.slack.com/archives/C0DLE125Q/p1645631159252259

Watch loom for more details:
https://www.loom.com/share/14a6cca013344049860b5f5367da3b80

### DIFF
--- a/cypress/fixtures/accountObFreeUpgrade.json
+++ b/cypress/fixtures/accountObFreeUpgrade.json
@@ -1,0 +1,629 @@
+{
+   "data":{
+      "account":{
+         "id":"606b65d6dad38c9d20b70792",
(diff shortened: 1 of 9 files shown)


cheahenry (approved):
LGTM!
Thanks for fixing it and adding a test!

PR #171, opened 2022-02-24
[GROW-393] Add Bugsnag to app-shell

cheahenry:
Created a new project in Bugsnag for [App Shell](https://app.bugsnag.com/buffer/app-shell/errors?filters[event.since]=30d&filters[error.status]=open&sort=last_seen)

This will enable Bugsnag only for `production` and `staging`.
This will also push errors to the Slack channel [#alerts-appshell](https://buffer.slack.com/archives/C034AEHFX9U/)

Tested locally, here's an [example of error](https://app.bugsnag.com/buffer/app-shell/errors/62176b77b1a02900082c4638?filters[event.since]=30d&filters[error.status]=open)

This is just a basic integration, there is more to do:
- Enable source maps so its easier to debug.
- Move API key to webpack environment var or somewhere else.

### DIFF
--- a/packages/app-shell/package.json
+++ b/packages/app-shell/package.json
@@ -64,6 +64,8 @@
     "@bufferapp/buffer-tracking-browser-ts": "^0.0.21",
     "@bufferapp/features": "^0.3.0",
     "@bufferapp/ui": "^5.75.7",
+    "@bugsnag/js": "^7.16.1",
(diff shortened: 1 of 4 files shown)


hamstu:
Thank you for starting this @cheahenry! All the Bugsnag links 404 for me though, I wonder if I need permissions on it?

PR #176, opened 2022-02-28
Set NODE_ENV for development

dace:
Sets `process.env.NODE_ENV` for development. 

We noticed that local development errors are alerting in Bugsnag ([Slack Thread](https://buffer.slack.com/archives/CPP95LDP1/p1646044554715609)). This sets  `development` as the environment when working locally to prevent these errors from reporting.


cheahenry (approved):
Nice, thanks for doing this!

PR #180, opened 2022-03-02
[COREP-610] Display email verification required banner

dace:
### Description

This PR:

- Adds a banner for users to re-send email verifications is their account returns with `shouldShowEmailVerificationCommunication`
- Adds a `dismissible` prop for the `Banner` component to remove the ability to dismiss a banner. For example, if a user were to dismiss the email verification banner, the expected behavior is that it would prevent it from appearing again, but as soon as the page refreshes they would see it. Hiding the ability to dismiss it makes it feel like a better UX in that scenario. The default is to allow dismissing, so you need to explicitly remove it. 
- Adds a check that an action is provided, otherwise don't show a button. 

### Jira
https://buffer.atlassian.net/browse/COREP-610

### Related PRs:
https://github.com/bufferapp/core-authentication-service/pull/1566
https://github.com/bufferapp/buffer-api-gateway/pull/415

### DIFF
--- a/cypress/fixtures/accountObEssential.json
+++ b/cypress/fixtures/accountObEssential.json
@@ -1,620 +1,609 @@
 {
-   "data":{
-      "account":{
-         "id":"606b65d6dad38c9d20b70792",
-         "email":"ob_essential@buffer.com",
-         "createdAt":"2021-04-05T19:32:38.841Z",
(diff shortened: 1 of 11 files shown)


philippemiguet (approved):
🚀

PR #181, opened 2022-03-02
[UPDATE] Allow plan update when only quantity is changed

daisymarie128:
🚢  This PR allows users to update plans with the new amount of channels but stay on the same plan.

Jira ticket: https://buffer.atlassian.net/browse/GROW-387

This includes new unit tests to cover updates to existing functions
Screen Shot 2022-03-02 at 2 45 31 pm

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
+++ b/packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
@@ -9,8 +9,10 @@ const useButtonOptions = ({
   hasPaymentDetails,
   isActiveTrial,
   isAwaitingUserAction,
+  currentChannelQuantity,
+  updatedChannelQuantity,
 }) => {
-  const getLabel = (selectedPlan) => {
+  const getLabel = (selectedPlan, updatedQuantity) => {
     if (isActiveTrial) {
       if (selectedPlan.planId === 'free') {
         return 'Confirm Plan Change';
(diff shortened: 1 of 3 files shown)


packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
```diff
@@ -9,22 +9,28 @@ const useButtonOptions = ({
...
         return 'Confirm Plan Change';
       } else return hasPaymentDetails ? 'Confirm Trial Plan' : 'Go To Payment';
     } else if (selectedPlan?.isCurrentPlan) {
       if (selectedPlan.planId === 'free' && isAwaitingUserAction) {
         return 'Confirm Free Plan';
+      } else if (currentChannelQuantity !== updatedQuantity) {
+        return 'Confirm changes';
       } else return 'Stay On My Current Plan';
     } else if (hasPaymentDetails || selectedPlan.planId === 'free') {
       return 'Confirm Plan Change';
```
cheahenry, line 27: I am not sure it's worth having different CTA copies if a plan change vs only quantity changes.
We could have all changes using the CTA `Confirm Changes`
See discussion [here](https://buffer.slack.com/archives/C0DLE125Q/p1645627368169019?thread_ts=1645626421.154559&cid=C0DLE125Q)

cheahenry (approved):
Logic LGTM, I pulled the PR and tested locally.

Added comments for CTA copy, not a blocker.

hamstu (approved):

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
daisymarie128: @cheahenry I like your suggestion here 👍 My plus 1 for this would also be to help simplify some of the logic in here. I feel this conditional statement is quite complex now. 
Also with the latest updates to the payment summary details which help describe which plan you're on and which you are changing too, I wonder how much extra value having different cta texts brings to users 🤔 

I will still merge this in now but would love to explore this further 😄

PR #183, opened 2022-03-03
Task/grow 390 messaging for channel input

daisymarie128:
This PR adds new ui messaging to the channel counter.

Jira ticket: https://buffer.atlassian.net/browse/GROW-390

Currently, still a draft as I haven't been able to test all the scenarios which could break the submit button with the new code added.

Screen Shot 2022-03-03 at 3 17 00 pm
Screen Shot 2022-03-03 at 3 16 40 pm

Screen Shot 2022-03-03 at 3 17 09 pm

Screen Shot 2022-03-03 at 3 17 17 pm

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js
+++ b/packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 
 import {
   getFreePlanChannelInputMessaging,
@@ -25,22 +25,6 @@ const useChannelsCounter = (
   function increaseCounter(disableIncrease = false) {
     if (!disableIncrease) {
       setChannelsCount(channelsCount + 1);
-
-      if (
-        planId === 'free' &&
-        channelsCount >= 3 &&
-        channelCountMessageStatus === null
-      ) {
-        setChannelCountMessageStatus(freePlanLimitMessageStatus);
-      }
-
-      if (
-        planId === 'agency' &&
-        channelsCount >= 10 &&
-        channelCountMessageStatus !== null
-      ) {
-        setChannelCountMessageStatus(null);
-      }
     }
   }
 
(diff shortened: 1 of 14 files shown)


packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js
```diff
@@ -10,20 +25,54 @@ const useChannelsCounter = (initialChannelCount = 0, minimumQuantity = 1) => {
...
+      planId === 'free' &&
+      channelsCount <= 4 &&
+      channelCountMessageStatus !== null
+    ) {
+      setChannelCountMessageStatus(null);
+    }
+
+    if (
+      planId === 'agency' &&
+      channelsCount === 10 &&
```
dinostheo, line 62: This will work but we should use the `minimunQuantity` of the `channelSlotDetails`, [see here](https://github.com/bufferapp/core-authentication-service/blob/main/core-api/src/graphql/schema.graphql#L348). In that way, we avoid hardcoding values and use a similar code, for every possible plan. Also, we have noticed in the past that we tend to experiment and change those values. It is less error-prone if we are having a single source of truth for those values. 

Regarding the free plan, we could also use the hard limit of the plan.

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js
```diff
@@ -10,20 +25,54 @@ const useChannelsCounter = (initialChannelCount = 0, minimumQuantity = 1) => {
...
   }
 
   function decreaseCounter(disableDecrease = false) {
     if (channelsCount > minimumQuantity || disableDecrease) {
       setChannelsCount(channelsCount - 1);
     }
+
+    if (
+      planId === 'free' &&
+      channelsCount <= 4 &&
```
dinostheo, line 54: Shouldn't this value be `< 4` or `<= 3`? 
Also I realize that we might want to add the `limits` on the plan options as well. https://github.com/bufferapp/core-authentication-service/blob/main/core-api/src/graphql/schema.graphql#L131

packages/app-shell/src/components/Modal/utils.js
```diff
@@ -114,6 +114,21 @@ export function calculateTotalSlotsPrice(
   return numberOfSlots * slotPrice;
 }
 
+export function getAgencyPlanMinimumChannelInputMessaging() {
+  return {
+    messageStatus: 'warning',
+    message: 'Oops! The Agency Plan has a minimum of 10 channels. ',
```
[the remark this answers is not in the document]
daisymarie128: Oh interesting 🤔 This is how it's setup to show now.

Have the Agency plan selected. 
When you have 10 channels selected and then try to click the - again it should show up.
It will then disappear if you increase the channel slots.

Though I'm not sure if this interaction is perfect because the message will still stay there if you leave it on 10 channels. 🤔

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js
daisymarie128: Nice one @dinostheo , I've updated this now, but I just wasn't sure about the Free plan. What did you mean to use the hard limit of the plan? Where would I find this value? I couldn’t find it in the plan details so have left it hard coded for now.

daisymarie128:
Screen Shot 2022-03-07 at 11 54 31 am
Screen Shot 2022-03-07 at 11 54 12 am

Latest updates addressing the comments above.
Also added some styling fixes to ensure the summary text doesn't flow behind the submit button.
All the code should now be behind the FF so I will merge this now :)

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js
dinostheo: @daisymarie128 this value is already available under the linked structure. However, I think it would be better for our use case if we would also add it to the channel details.

PR #185, opened 2022-03-07
Fix/z index modals

daisymarie128:
Link to bug reported on prod: https://buffer.slack.com/archives/C0DLE125Q/p1646170700283989

The navigator has recently been forced on top of everything for upcoming work to do with paywall and nondismissable modals. 

To fix this for now we have added a min height to the ModalWrapper so users on a smaller screen can scroll the nav bar away.

**Testing notes:**
Tested on a few different small screen sizes and I seem to be able to scroll the nav bar away.
@hamstu did you want to test against the scenario you originally encountered with? If not I'll merge this in tomorrow 😊

Screen Shot 2022-03-07 at 4 44 55 pm
Screen Shot 2022-03-07 at 4 44 46 pm


federicoweber (approved):
Thanks @daisymarie128 🙌

PR #186, opened 2022-03-07
[GROW-405] Update Confirm Trial Plan to Confirm Plan to be less confusing

cheahenry:
Updating some copy and a CTA to be less confusing from `Confirm Trial Plan` to `Confirm Plan`.

Screenshot 2022-03-07 at 16 38 09


dinostheo (approved):
🚢

PR #187, opened 2022-03-07
Update e-mail verification banner copy to provide more clarity for users

dace:
Updates the email verification banner copy for better UX:

CleanShot 2022-03-07 at 10 27 03@2x

Related Slack thread: https://buffer.slack.com/archives/C6JL89JG6/p1646660985383169?thread_ts=1646655318.471679&cid=C6JL89JG6

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -2,6 +2,5 @@
   "video": false,
   "viewportWidth": 1440,
   "viewportHeight": 900,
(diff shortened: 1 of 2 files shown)


dace:
Merging this with failing Cypress tests as the Growth team is aware and looking into it. 

Related Slack thread: https://buffer.slack.com/archives/C0DLE125Q/p1646669562408999

PR #190, opened 2022-03-08
[GROW-384] Standalone channel quantity modal

gisete:
Standalone modal that allows user to add/remove channels

image


packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
```diff
@@ -0,0 +1,143 @@
...
+          </InnerContainer>
+        </Section>
+      </SectionContainer>
+
+      <ButtonWrapper>
+        <Button type="text" onClick={() => openModal(null)} label="Cancel" />
+        <Button
+          type="primary"
+          onClick={(data) => {
+            openModal(MODALS.paymentMethod, data);
```
dinostheo, line 134: I notice that we are opening the payment method modal. Is there a flow that we will be asking the credit card details of a customer from the Quantity modal? I believe that we should already have this information, since the customer needs to be on a paying plan to work with this modal.

packages/app-shell/src/components/Modal/modals/QuantityUpdate/index.jsx
```diff
@@ -36,58 +22,32 @@ const QuantityUpdate = () => {
...
+            const {
+              flatFee,
+              currentQuantity,
+              pricePerQuantity,
+              minimumQuantity,
+            } = user.currentOrganization.billing.channelSlotDetails;
+
+            const { name: planName, id: planId } =
+              user.currentOrganization.billing.subscription.plan;
+            const planOptions =
```
dinostheo, line 34: We shouldn't need to dig the current plan from the `changePlanOptions`. I believe we should have all the information you need under that [subscription.plan and the plan.prices](https://github.com/bufferapp/core-authentication-service/blob/main/core-api/src/graphql/schema.graphql#L390-L427). 
If something is missing we should better add it to this structure instead of using the plan options.

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
```diff
@@ -0,0 +1,143 @@
...
+                  onIncreaseCounter={() => increaseCounter()}
+                />
+              </ChannelCounterWrapper>
+            </div>
+
+            <Summary>
+              {hasCounterChanged ? (
+                <>
+                  <Text type="p">
+                    New monthly cost: <strong>${newPrice}</strong>
```
dinostheo, line 113: It feels a bit that the 2 texts here might be conflicting `New monthly cost` and below `This will be billed every {planCycle} until canceled`, with `planCycle` to be either month or year. 
I'm might be understanding wrong though 😅

dinostheo (changes requested):
This already looks great @gisete. It's great that you have merged since it's behind the feature flip. 
I have posted some questions and some remarks, with the latter needing some changes.

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
cheahenry: yeah I think we can just keep `New cost` and keep the planCycle in the text below like we do in the plan selector

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
```diff
@@ -0,0 +1,143 @@
...
+  planCycle,
+  openModal,
+}) => {
+  const [hasCounterChanged, changeCounterState] = useState(false);
+
+  const {
+    channelsCount,
+    // setChannelsCounterValue,
+    increaseCounter,
+    decreaseCounter,
```
cheahenry, line 46: You can probably also use `channelCountMessageStatus` for message status
https://github.com/bufferapp/app-shell/blob/2fc33dcfd7abcac04d4b05ead254aa2e088e993e/packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useChannelsCounter.js#L88

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
```diff
@@ -0,0 +1,143 @@
...
+        <Text type="p">
+          You&apos;re currently on the <strong>{planName}</strong> plan and
+          you&apos;re paying{' '}
+          <strong>{getProductPriceCycleText(planPrice, planCycle)}</strong> for{' '}
+          {quantity} channel
+          {quantity !== 1 ? 's' : ''}.{' '}
+          <button
+            type="button"
+            onClick={(data) => {
+              openModal(MODALS.planSelector, data);
```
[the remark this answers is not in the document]
dinostheo: Thanks @gisete. 
Now I can picture what is happening here. I believe it is from the following design 
Screen Shot 2022-03-10 at 17 08 41

and it feels that the intention is indeed, on `Change Plan` CTA to open the Plan Selector and change plan. In that case, your implementation is correct. 
Regarding if this is a good UX flow, I cannot really tell 😅 . I believe you were spot on implementing this design though 😄

gisete:
> 

Good tip @dinostheo. I haven't been intentional with squashing my commits before merging but I like when repos have that by default because it doesn't let me forget 🙂  I'll look into how I can turn it on for this repo.

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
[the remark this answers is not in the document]
gisete: @dinostheo I apologize, I thought you were talking about the action to the "Confirm & Pay" button, ignore my comment. Indeed this is meant to open the Plan Selector according to the design and I think it makes sense here. This copy and action is similar to the CTA that was added to the Channels page.

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
gisete: @dinostheo I created a sub-task in Jira for us to update the action here. I don't know yet how to go make a payment go through. Do you have an example in the code where that's being done?

dinostheo: @gisete In the following, we call the success modal https://github.com/bufferapp/app-shell/blob/main/packages/app-shell/src/components/Modal/modals/PlanSelector/index.jsx#L42
However, we need to do that after we have made a request to the server-side to persist the quantity change. We have a mutation for it on the API, [billingUpdateSubscriptionQuantity](https://github.com/bufferapp/core-authentication-service/blob/main/core-api/src/graphql/schema.graphql#L813-L817). We will need to create a mutation on the client-side to call this one and handle the success response by opening the success modal (or the error). 

[Here's how we make the API call](https://github.com/bufferapp/app-shell/blob/278ac9a29312883de572f79014d42742b0a2a83b/packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx#L111-L122) in the case of the Plan Selector (it's a different mutation here)
and here's [how we handle the response](https://github.com/bufferapp/app-shell/blob/278ac9a29312883de572f79014d42742b0a2a83b/packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx#L213-L224)

All these might be very new due to GraphQL, so happy to pair on it, on Monday or any other day next week 😃

packages/app-shell/src/components/Modal/modals/QuantityUpdate/index.jsx
gisete: @dinostheo I created a PR to address this https://github.com/bufferapp/app-shell/pull/201

PR #195, opened 2022-03-10
Add AppShell Composer [Webapp-385] #RFC

federicoweber:
# Introducing AppShell Composer
The `Composer` is a new player in our Micro Frontends strategy that allows us to render **components, from AppShell, in other Product Apps**, using similar APIs to the one used to open a modal.

```javascript
const { actions, COMPONENTS } = window.appshell
actions.renderComponent({ 
  componentKey: COMPONENTS.startTrialButton, // The component key
  containerId: 'portal1', // The ID of the DOM node we are targetting
  options: {...}, // A collection of options to pass down to the component
})
```

Orchestrator

## Why
Working on [WEBAPP-385](https://buffer.atlassian.net/browse/WEBAPP-385), I needed to find a good way to directly start the trial from a CTA without opening the StartTrial modal. With the current setup, this isn't trivial, because we don't have a good way to share business logic between `AppShell` and the rest of the apps.

## Benefits
- easily reusable components across all apps
- full encapsulation of the components (context and libraries are separated)
- automatic update across all apps (same process as the `Navigator`
- components have access to the contexts available in the `AppShell`
- events from a component will bubble up all the to the `AppShell`

## Cons
- It's not possible to compose components in this way, as it will cross two different React apps

## How does this work
This is possible thanks to the use of [React.Portals](https://reactjs.org/docs/portals.html). Effectively this is quite similar to how the `ReactDOM.render` works, but instead of forcing the application to be nested in a single DOM node, it allows us to borrow nodes from across the entire Document.


packages/app-shell/src/common/events/orchestratorEvents.js
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const ACTIONS = {
+  renderComponent({ componentKey, containerId, options }) {
+    eventDispatcher(EVENT_KEY, { action: ACTION_KEYS.renderComponent, componentKey, containerId, options });
+  },
+  unmountComponent({ containerId }) {
+    eventDispatcher(EVENT_KEY, { action: ACTION_KEYS.unmountComponent, containerId });
+  },
+};
```
federicoweber, line 17: This is still unused at the moment

packages/app-shell/src/common/events/orgEvents.js
```diff
@@ -0,0 +1,23 @@
...
+      EVENT_KEY,
+      { action: ACTION_KEYS.setCurrentOrganization, organizationId, options }
+    )
+  },
+  currentOrganizationUpdated(options = {}){
+    eventDispatcher(
+      EVENT_KEY,
+      { action: ACTION_KEYS.currentOrganizationUpdated,  options }
+    )
+  }
```
federicoweber, line 22: I've added this event so that we have a way for the apps to know when something changed in the organization, for example, if the plan is upgraded, and react accordingly.

packages/app-shell/src/exports/Navigator/index.jsx
```diff
@@ -120,6 +126,7 @@ export const Navigator = React.memo(({ apolloClient, channels }) => {
         )}
         {showEmailVerificationBanner && <EmailVerificationBanner />}
         {!user.loading && <Modal {...modal} />}
+        <Orchestrator />
```
federicoweber, line 129: We are rendering the `Orchestrator` here so that its components have access to all the available contexts.

packages/app/src/App.js
```diff
@@ -60,16 +79,19 @@ const ModalTesting = () => {
...
-          actions.openModal(MODALS.paywall, {
-            cta: 'paywall',
-          });
+          const { COMPONENTS, actions } = window?.appshell || {};
+          actions.renderComponent({
+            componentKey: COMPONENTS.startTrialButton,
+            containerId: 'components_container',
+            cta: 'orchestrator_trial',
+            ctaButton: 'orchestrator_button'
+          })
```
federicoweber, line 91: This is a quick example on how we can render a component in response to a user action.
If we want to append a component when it's parent render we can do that with an effect

```javascript
const Foo = () => {
  useEffetc(() => {
   const { COMPONENTS, actions } = window?.appshell || {};
   actions.renderComponent({containerId: 'foo', ...})
  }, [])
  return (<div id='foo' />)
}
```

federicoweber:
You can take a look at a use case for this in https://github.com/bufferapp/buffer-account/pull/758 :)

packages/app-shell/src/components/Modal/modals/Paywall/components/Ctas.jsx
```diff
@@ -4,7 +4,7 @@ import Link from '@bufferapp/ui/Link';
 import Text from '@bufferapp/ui/Text';
 
 import useStartTrial from '../../../../../common/hooks/useStartTrial';
-import { useSuggestedPlan } from '../../../../../common/hooks/useSuggestedPlan';
+import { getSuggestesPlan } from '../../../../../common/utils/getSuggestedPlan';
```
msanroman, line 7:
```suggestion
import { getSuggestedPlan } from '../../../../../common/utils/getSuggestedPlan';
```

packages/app-shell/src/components/Modal/modals/Paywall/components/Ctas.jsx
```diff
@@ -19,7 +19,7 @@ const Content = ({ openModal }) => {
   const canStartTrial = user?.currentOrganization?.billing?.canStartTrial;
   const cta = canStartTrial ? 'startTrial' : 'upgradePlan';
 
-  const { suggestedPlan } = useSuggestedPlan(user);
+  const suggestedPlan = getSuggestesPlan(user);
```
msanroman, line 22:
```suggestion
  const suggestedPlan = getSuggestedPlan(user);
```

packages/app-shell/src/components/Modal/modals/StartTrial/index.jsx
```diff
@@ -9,13 +9,13 @@ import { MODALS } from '../../../../common/hooks/useModal';
 import { UserContext } from '../../../../common/context/User';
 import { ModalContext } from '../../../../common/context/Modal';
 import useStartTrial from '../../../../common/hooks/useStartTrial';
-import { useSuggestedPlan } from '../../../../common/hooks/useSuggestedPlan';
+import { getSuggestesPlan } from '../../../../common/utils/getSuggestedPlan';
```
msanroman, line 12:
```suggestion
import { getSuggestedPlan } from '../../../../common/utils/getSuggestedPlan';
```

packages/app-shell/src/components/Modal/modals/StartTrial/index.jsx
```diff
@@ -9,13 +9,13 @@ import { MODALS } from '../../../../common/hooks/useModal';
...
 import { useTrackPageViewed } from '../../../../common/hooks/useSegmentTracking';
 
 import { Holder, Content, Ctas } from './style';
 
 const StartTrial = ({ user, openModal, modalData }) => {
-  const { suggestedPlan } = useSuggestedPlan(user);
+  const suggestedPlan = getSuggestesPlan(user);
```
msanroman, line 18:
```suggestion
  const suggestedPlan = getSuggestedPlan(user);
```

packages/app-shell/src/exports/Navigator/index.jsx
```diff
@@ -63,6 +65,10 @@ export const Navigator = React.memo(({ apolloClient, channels }) => {
         };
   useUserTracker(user);
 
+  useEffect(() => {
+    ORGANIZATION_ACTIONS.currentOrganizationUpdated({ user });
```
msanroman, line 69: We might want this to be triggered only when the organization was actually updated, right? In this scenario it might get triggered more often 🤔

packages/app-shell/src/exports/Orchestrator/components/StartTrialButton.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { useEffect } from 'react';
+import Button from '@bufferapp/ui/Button';
+
+import useStartTrial from 'common/hooks/useStartTrial';
+import { getSuggestesPlan } from 'utils/getSuggestedPlan';
```
msanroman, line 5:
```suggestion
import { getSuggestedPlan } from 'utils/getSuggestedPlan';
```

msanroman (approved):
I don't have any blocking concerns, this looks good to me. I left some quick renaming comments to fix a typo and some question about the use of `useRef`.

The main reflection I'd like to leave here is whether this approach and developer experience feels ergonomic for our codebases or if there's a different approach to take here that would make it easier for anyone to consume. Maybe we don't need all kinds of components to be as up-to-date in every app as we were challenged with the AppShell integration.

If that's the case, we could have some components like a `StartTrialButton` with self-contained logic in our React component library. Or maybe we should dig more into whether a monorepo approach would benefit the frontend too. These are just quick thoughts, nothing too developed in my mind!

This approach looks solid to me, but wanted to leave some reflections here too! 😄

packages/app-shell/src/exports/Orchestrator/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+      return null;
+  }
+}
+
+export const Orchestrator = () => {
+  const [components, setComponents] = useState([])
+  const componentsRef = useRef()
+
+  componentsRef.current = components
+
```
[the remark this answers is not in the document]
federicoweber: Good question @msanroman.
I'm using refs in here so that the `handleEvents` can access the update `components` value. Otherwise, it will only access the value that has been provided on initialization, an empty array, and we won't be able to render more than one component 🙂.

packages/app-shell/src/exports/Navigator/index.jsx
federicoweber: Good one, I was considering doing a deep comparison in here, but I think it won't be necessary, as the only nonuser actions-related case that comes to mind for this object to change is on load, and in that case.

Do you think there could be other cases for this to fire?

msanroman: It doesn't right now, but it might in the future (other attributes that mutate on runtime), and this could cause unexpected side effects. I'd rather have us encode this business logic more explicitly: we want to trigger this event if the user goes from `loading` to `loaded` and if the `currentOrganization.id` changes (I'm making up the values, it's been a while 😅 )

If that's the case, I'd change the dependencies in the `useEffect` call to be `[user?.currentOrganization?.id]`, that should go from `null` to the loaded `id`, and trigger again whenever the id changes. Or something like that! Again, it's been a while since I looked into the user object 😂

federicoweber:
Thanks for the review @msanroman, great catch on the typo 🙌

> The main reflection I'd like to leave here is whether this approach and developer experience feels ergonomic for our codebases or if there's a different approach to take here that would make it easier for anyone to consume. Maybe we don't need all kinds of components to be as up-to-date in every app as we were challenged with the AppShell integration.

> If that's the case, we could have some components like a StartTrialButton with self-contained logic in our React component library.

Totally agree, I don't think this is a catch-all solution, and I would suggest limiting this to mission-critical components, that need to always be up to date.

> Or maybe we should dig more into whether a monorepo approach would benefit the frontend too. These are just quick thoughts, nothing too developed in my mind!
Yess, will share some notes in a thread later, but we officially started the exploration for the front-end mono repo within the FE-WG 😁. Once we get to that we should also be able to sunset both the components and the modals APIs, from the Navigator.

packages/app-shell/src/exports/Navigator/index.jsx
federicoweber: The main use case for this specific event is when the billing information changes, for example after a trial start, so the host App can properly refetch the data.

I can look into limiting the scope of this event, to just that case. :)

federicoweber: Hey @msanroman, I updated the logic so that we are dispatching a `billingUpdated` action, only when we mutate the billing.

PR #198, opened 2022-03-13
[COREP-617] Prevent user from updating payment method if email is unverified

dace:
Displays a notice and prevents the user from updating payment method if they have not verified their email address

Before email verification:

https://user-images.githubusercontent.com/10112294/158360935-d9315691-5718-4f62-bd38-a2dcb7055b2d.mp4

After email verification:

https://user-images.githubusercontent.com/10112294/158361509-e6f06ac5-36e2-4c3c-913c-3290479c4522.mp4

### Jira
[COREP-617](https://buffer.atlassian.net/browse/COREP-617)


dace:
@daisymarie128 I'm noticing that Cypress is failing again for this PR. It looks similar to the error from last week I encountered that wasn't very descriptive. Is there a way to better understand what the actual cause of Cypress errors is? I believe you solved the last error by adding a missing property in a Cypress fixture file. I checked and they all seems to be fine, so just wondering if maybe there's a secret to recognizing the real issue 😄

daisymarie128:
@dace I see your pipeline is green now which is good, sorry about the failing Cypress test though, did you manage to figure out what was causing it to break?

If not I know there is one flakey test that was failing most times, the error was something along the lines of `cannot find #InviteCta` - Re-running the tests seems to fix it though 🤦‍♀️ But I haven't been able to figure out why it only fails sometimes 😢 .

dinostheo (approved):
LGTM @dace 🙌

PR #202, opened 2022-03-14
[GROW - 422] Disable submit button if # in counter hasn't changed

gisete:
Disabled "Confirm and Pay" submit button if number of channels in the counter is the same as the number of channels the customer already has.

Also cleaned up the code a bit.

image


daisymarie128 (approved):
🎉 I think because I merged the agency restriction PR there are conflicts now sorry @gisete 🙏

gisete:
No worries @daisymarie128 ! I see you already have a solution for this in your branch so I'll wait until you merge and just updated the code to include the small refactor I pushed.

PR #203, opened 2022-03-15
Task/grow 421 action for confirm btn

daisymarie128:
Jira ticket: https://buffer.atlassian.net/browse/GROW-421

This PR hooks up the functionality to make the "Confirm & Pay" button work in the QuantityUpdate modal

This pr also includes:
- new actions file which holds all "actions" functions that fire on events. (This was done to try and minify how much logic gets put into components themselves and hopefully make the code more readable. 
- Unit tests
- Refactor ChannelsCounter to useEffect and listen for changes originally being done in the components. 
- disables "Confirm & pay" until user changes channel counter
Screen Shot 2022-03-15 at 3 10 25 pm


packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
```diff
@@ -115,16 +148,23 @@ const CardBody = ({
             </Summary>
           </InnerContainer>
         </Section>
+        {error && <Error error={error}>{error.message}</Error>}
```
daisymarie128, line 151: 👀 I added this just following a pattern which is used in the plan selector.
But I haven't actually tested this and seen what it looks like yet, because I wasn't sure how to make an actual error come back. 

I will try hard code something and force and error tomorrow, unless anyone knows a better way to manually test this one 😅

packages/app-shell/src/components/Modal/modals/QuantityUpdate/hooks/useButtonOptions.js
```diff
@@ -0,0 +1,31 @@
+import { useState, useEffect } from 'react';
+
+function useButtonOptions({
```
daisymarie128, line 3: This follows a similar method that the PlanSelector uses to determine which screen to show.

This will handle either opening the Credit card payment modal or process the payment using the saved payment method

packages/app-shell/src/components/Modal/modals/QuantityUpdate/actions.js
```diff
@@ -0,0 +1,44 @@
+import { MODALS } from '../../../../common/hooks/useModal';
+
+export function onSuccess(plan, openModal) {
+  if (!plan || !openModal) {
```
daisymarie128, line 4: I added these generic console errors due to the fact while working on this ticket I noticed currently when you try and open modals there wasn't any helpful messaging letting you know why something wasn't working.

Ideally in the future, I feel like:
1. Would be best we throw actual errors - went with this just as a bare minimum to help debug things
2. The error handling should get added to the Modal index or a higher level of the application instead 🤔

packages/app-shell/src/common/hooks/useUpdateSubscriptionPlan.js
```diff
@@ -22,12 +22,16 @@ const useUpdateSubscriptionPlan = ({
...
         ORGANIZATION_ACTIONS.billingUpdated({ user });
-      }
+      },
     }
   );
 
   useEffect(() => {
     if (!user || !plan || !hasPaymentMethod) {
+      // eslint-disable-next-line no-console
+      console.error(
```
daisymarie128, line 32: I added this because when none of these things are present atm it silently fails and took a bit of time to find exactly what was causing the modals to get stuck in a processing state.

I didn't want to throw an error just yet as I haven't properly considered how that could affect current user flows and behaviors.

Do let me know if we would all prefer I just removed these new console errors I've added in this PR or if there's a preferred way 🙂

packages/app-shell/src/common/hooks/useChannelsCounter.js
```diff
@@ -78,6 +88,11 @@ const useChannelsCounter = (
 
   useEffect(() => {
     handleWarningMessageChecks();
+    handleChannelsCountConditions(
```
daisymarie128, line 91: This part of the code isn't actually related to the ticket. But I saw we could simplify some of the code and listen to this bit of code that was already written and remove it from components.

packages/app-shell/src/common/hooks/useUpdateSubscriptionPlan.js
dinostheo: @daisymarie128 ideally we would want to use its dedicated mutation for the Quantity modal. 
[billingUpdateSubscriptionQuantity](https://github.com/bufferapp/core-authentication-service/blob/main/core-api/src/graphql/schema.graphql#L819-L824)
It is a simpler mutation on the call side, as it does not require the plan information (planId and interval). And also it is way simpler on the server side, as it has a single API call to Stripe instead of a multi-step logic that the update plan has. The reason is that only customers on a paid plan can purchase channels and that means that we don't have to handle their payment method, instantiate the Stripe form with an intend ID and proactivelly remove any active trials or cancelations (cancel at period end).

Long story short the above (new) mutation only increases the quantity of the subscription, which is actually what we need. 

That said we don't need to check for payment details either. Organizations with a paid plan, have already set that.

packages/app-shell/src/components/Modal/modals/QuantityUpdate/actions.js
dinostheo: I'd love to see that 😍

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
```diff
@@ -58,12 +90,13 @@ const CardBody = ({
   });
 
   useEffect(() => {
-    handleChannelsCountConditions(
-      planId,
-      channelsCount,
-      setChannelsCounterValue
-    );
-  }, [channelsCount]);
+    if (updatePlanData?.billingUpdateSubscriptionPlan.success) {
```
dinostheo, line 93: Once we change the mutation we will have to change the success handling as well to use the [BillingUpdateSubscriptionQuantityPayload](https://github.com/bufferapp/core-authentication-service/blob/main/core-api/src/graphql/schema.graphql#L733)

packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx
dinostheo: I believe we don't need a credit card form here. It adds too much complexity in the flow and we already have the credit card information of the customer, since only paid customers can purchase channels.

dinostheo (changes requested):
That's great work @daisymarie128 Thanks for adding all the error logs to bring more visibility on what is happening here 😄 

We need a major change here though, which is the mutation that we should call. The Quantity modal should not use the same mutation with the Plan Selector to update Billing. The reason is that we already have the credit card of the customer and we don't need to rehandle this complexity. On top of that, the quantity mutation makes fewer API calls to STripe, which makes it faster as well as less prone to API call failures. 
I believe I didn't communicate this mutation well though, so that's on me 😅 .

daisymarie128:
> That's great work @daisymarie128 Thanks for adding all the error logs to bring more visibility on what is happening here 😄
> 
> We need a major change here though, which is the mutation that we should call. The Quantity modal should not use the same mutation with the Plan Selector to update Billing. The reason is that we already have the credit card of the customer and we don't need to rehandle this complexity. On top of that, the quantity mutation makes fewer API calls to STripe, which makes it faster as well as less prone to API call failures. I believe I didn't communicate this mutation well though, so that's on me 😅 .

@dinostheo No sorry, I honestly just picked up the ticket then went straight into coding mode 😅 I should have gone and read more context on this modal, so that's on me too 🙏 
But thank you for flagging this and sharing more context here 🌮 This way you explain makes so much sense and a much better approach I love it!

dinostheo (approved):
Thanks for doing all this cleanup 🙏

PR #201, opened 2022-03-14
Data source update and code cleanup

gisete:
Following [Dinos' advice here](https://github.com/bufferapp/app-shell/pull/190#discussion_r823778103), I updated what properties we're using to populate the customer's plan price and cycle.

PR #205, opened 2022-03-15
[GROW-428] use totalPrice instead of basePrice for description

cheahenry:
`planPrice` is only used in the description
https://github.com/bufferapp/app-shell/blob/4b48ab7df08617be2ef27a11e06664cab71e7ca6/packages/app-shell/src/components/Modal/modals/QuantityUpdate/components/CardBody.jsx#L72-L77

Before
Screenshot 2022-03-15 at 11 19 17

After
Screenshot 2022-03-15 at 11 18 44

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/QuantityUpdate/index.jsx
+++ b/packages/app-shell/src/components/Modal/modals/QuantityUpdate/index.jsx
@@ -35,7 +35,7 @@ const QuantityUpdate = () => {
             const planOptions =
               user.currentOrganization.billing.changePlanOptions;
             const currentPlan = getCurrentPlanFromPlanOptions(planOptions);
-            const { basePrice: planPricing, planInterval } = currentPlan;
+            const { totalPrice: planPricing, planInterval } = currentPlan;
             return (
               <Container>
                 <CardBody


packages/app-shell/src/components/Modal/modals/QuantityUpdate/index.jsx
```diff
@@ -35,7 +35,7 @@ const QuantityUpdate = () => {
             const planOptions =
               user.currentOrganization.billing.changePlanOptions;
             const currentPlan = getCurrentPlanFromPlanOptions(planOptions);
-            const { basePrice: planPricing, planInterval } = currentPlan;
+            const { totalPrice: planPricing, planInterval } = currentPlan;
```
gisete, line 38: @cheahenry Dinos suggested we don't use `changePlanOptions` to get the price data, I had created a PR to update it here https://github.com/bufferapp/app-shell/pull/201. Does `baseMonthlyPrice` hold the same value as `totalPrice`?

cheahenry: Hey @gisete,

Good call! I don't think we store the `totalPrice` in the datasource but it can be calculated using this function
https://github.com/bufferapp/app-shell/blob/b7f5f5409ebe7379e0fbdb30e7f7003aa5975fc0/packages/app-shell/src/components/Modal/utils.js#L97-L114

cc @dinostheo

dinostheo: @cheahenry @gisete total price will include the quantity as it is set in Stripe and not based on the selection on the Quantity modal. 
I believe we want to use the plan price that is calculated for the interval (not the monthly one), but without the quantity. 
If the above is correct we should use one of the following
[PlanPrices.basePlanPrice](https://github.com/bufferapp/core-authentication-service/blob/f6e808f0adac540aa12a5f2ac3a9ca0f02b053db/core-api/src/graphql/schema.graphql#L408)
[channelSlotDetails.pricePerQuantity](https://github.com/bufferapp/core-authentication-service/blob/f6e808f0adac540aa12a5f2ac3a9ca0f02b053db/core-api/src/graphql/schema.graphql#L348)

PR #208, opened 2022-03-15
[GROW-430] Fix the payment method summary for SBB

dinostheo:
[[GROW-430](https://buffer.atlassian.net/browse/GROW-430)] 
This PR is fixing 2 issues regarding SBB and the payment method modal. 

 1. Customers without a credit card can buy more channels via the plan selector when they don’t change plans. 
 2. Customers without a credit card cannot submit their credit card information.
 
I'm not sure about the approach and if it is following good react patterns. It feels a bit convoluted and I don't have the knowledge necessary to refactor those with confidence so close to the launch. 
It follows a similar pattern we currently use and it is functional 😅. 

However, be critical since it is the method customers pay with 😄 .


daisymarie128 (approved):
Nice one @dinostheo thanks for fixing this 🙌

packages/app-shell/src/components/Modal/modals/Summary/index.jsx
```diff
@@ -60,6 +62,28 @@ function renderSBBSummary(
...
   const selectedPlanUsersText = findPlanUserDetails(selectePlanSummary.details);
 
+  const paymentInfo = isPaymentMethodSummary ? (
+    <PaymentPlanInfo
+      planName={selectedPlanName}
+      planCycle={selectedPlanInterval}
+      numberOfUsers={selectedPlanUsersText}
+      channelsCount={channelsCount}
+      newPrice={newPrice}
+      channelCounterMessageStatus={channelCounterMessageStatus}
```
daisymarie128, line 72: I don't think you use this prop so it should get deleted 🙂

packages/app-shell/src/components/Modal/modals/PlanSelector/index.jsx
```diff
@@ -27,7 +27,10 @@ const PlanSelector = () => {
                 }
                 user={user}
                 openPaymentMethod={(data) => {
-                  modal.openModal(MODALS.paymentMethod, data);
+                  modal.openModal(MODALS.paymentMethod, {
+                    ...data,
```
daisymarie128, line 31: non-blocking: I find passing through whole objects using the spread can lead to lack of context. 
i.e 
```{
   ...data, // Apon reading it I have no idea what is being passed back and makes it harder to debug, while potentially passing back things we don't need.
}```

Compared to this pattern which is more strict about what it is that will be passed back. 
```
{
   plan: modal.data.plan,
   newPrice: modal.data.newPrice
}```

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
```diff
@@ -41,17 +41,12 @@ const useButtonOptions = ({
...
     }
 
-    if (
-      selectedPlan.isCurrentPlan &&
-      currentChannelQuantity !== updatedChannelQuantity
-    ) {
-      return updatePlan;
-    }
-
-    if (selectedPlan.isCurrentPlan && !isActiveTrial) {
```
daisymarie128, line 51: @dinostheo I don't know much about this property `isActiveTrial` . Just want to double-check that if they are on an active trial and they are on their current plan it doesn't need to return null anymore?

dinostheo: @daisymarie128 
`selectedPlan.isCurrentPlan && !isActiveTrial` would return `true` for the ones that are not on trial and haven't switched plans. However, we are already guarding for that on an earlier step and for SBB, this could mean that the customer might be purchasing or removing slots.

packages/app-shell/src/components/Modal/modals/Summary/index.jsx
dinostheo: I do think I need them for the data in the summary to render. If I remove them the `renderSBBSummary` method won't work. 
Let me know if I'm doing something wrong in this part.

packages/app-shell/src/components/Modal/modals/PlanSelector/index.jsx
dinostheo: I'm not sure about the existing data passed (the first spread), but I can make explicit the second "modalData" spread.

PR #213, opened 2022-03-17
fix: use AccountOrganizationFiels fragment to prevent missing props

esclapes:
We do some manual cache changes when an organization is switched in
order to prevent over-fetching but if the fields in both
currentOrganization and organizations are not the same the query
write will fail to validate and break.

Using a fragment ensures that we request the same fields on both edges.


packages/app-shell/src/common/hooks/useOrgSwitcher.js
```diff
@@ -20,6 +20,7 @@ function useOrgSwitcher() {
         currentOrganization: organizationSelected,
       },
     };
+
     client.writeQuery({
```
esclapes, line 24: This is where the error was happening.

PR #215, opened 2022-03-21
[FIX] Modal hidden under nav

daisymarie128:
Jira: [GROW-437](https://buffer.atlassian.net/browse/GROW-437)

Hacky fix but now the modal can be scrolled into view

👀  Would I be able to get someone to review this PR just as I tried many many different ways to fix this issue (including looking into making changes directly into the UI modal) I came up with this solution and used the awkward CSS child selectors to hopefully reduce the risk of breaking modal layouts throughout our applications 🤔  

Short loom video showing the bug and the new behaviour: https://www.loom.com/share/6d7adc275440463796e25b47cd979111?sharedAppSource=personal_library

### DIFF
--- a/packages/app-shell/src/components/Modal/index.jsx
+++ b/packages/app-shell/src/components/Modal/index.jsx
@@ -22,11 +22,38 @@ import TrialExpired from './modals/TrialExpired';
 import QuantityUpdate from './modals/QuantityUpdate';
 import { Paywall } from './modals/Paywall';
 
-import { shouldShowFreeUserStartTrialPrompt, shouldShowChannelConnectionPrompt, shouldShowPaywallModal } from './utils';
+import {
+  shouldShowFreeUserStartTrialPrompt,
+  shouldShowChannelConnectionPrompt,
+  shouldShowPaywallModal,
+} from './utils';
 
 const ModalWrapper = styled.div`
   > div {
     z-index: 1;
+    overflow: scroll;
+
+    // Hacky: The navigator has been forced on top of everything
+    // This is nbeeded for the paywall work.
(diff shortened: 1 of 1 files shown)


packages/app-shell/src/components/Modal/index.jsx
```diff
@@ -22,11 +22,38 @@ import TrialExpired from './modals/TrialExpired';
...
+        position: absolute;
+        bottom: -300px;
+        left: 0;
+        width: 100px;
+        height: 400px;
+      }
+
+      > div > div {
+        top: 20px;
+      }
```
gisete, line 55: @daisymarie128 it's hacky but it works! 😄   

It would be better if you didn't have to use this but since it's wrapped in a media query I think there are very low chances of this change breaking anything. I don't have a ton of context on where all the modals are being used though so it's just my take based on this PR.

gisete (commented):
I think this works but could we look into making this fix permanent by adding it to the Modal component instead? I think it would be helpful to get more eyes on it and we could run more tests to make sure it behaves the way it's expected.

federicoweber (changes requested):
Hey @daisymarie128 I would love to understand this a bit better.
Specifically, what are the reasons we are forcing this in here, instead of the UI library?

My concerns are that while this work it's very specific, and introduces a lot of coupling between the way the Modals are defined in the UI library, and how we are using those here.
Also, we will likely be facing similar issues in the future, when using the modals outside of AppShell.
I also have some hesitation about forcing the modal on top of the navigation bar, as it might prevent the user from any further interactions with Buffer, in cases of nondismissable modals.

Please let me know if you want to jump on a call and explore different alternatives, always happy to pair 😊.

PR #216, opened 2022-03-21
Fix opening the payment method modal for active trialists on the same plan

dinostheo:
This is fixing a bug, setting the inactive payment method action for trialists in their current plan without quantity change. 

See the screen recordings and the explanation from @dace in the following slack thread. 
https://buffer.slack.com/archives/C0DLE125Q/p1647865540724829


daisymarie128:
@dinostheo Thanks for fixing this plus adding the new tests to cover this scenario as well 🙏

PR #219, opened 2022-03-22
[GROW-440] remove top border summary

cheahenry:
`slot-based-billing` OFF
SBB OFF

`slot-based-billing` ON - before
SBB ON Before

`slot-based-billing` ON - after
Screenshot 2022-03-22 at 16 07 50


dinostheo (approved):
🚢

PR #224, opened 2022-03-23
[GROW-441] Decide on quantity update or plan selector based on the CTA and organization state

dinostheo:
We seem to have multiple upgrade CTAs, which should bring up the Quantity Update modal or the Plan Selector based on the Organization State. 
In some of the product cases such as Publish, we don't even have the billing information available at the place to make the decision easily and we might have to add very specific logic to every product. 

With this solution, we are trying to let the decision to the navigator, when the decision flag `shouldPickModalOnOrganizationState` is set.


cheahenry (approved):
Loving this, it was getting painful to have the same logic for all the upgrade CTAs!
Thanks for also adding tests!

cheahenry:
We can probably do something similar for locked members and `canEdit` flag ([ticket](https://buffer.atlassian.net/browse/GROW-429))

dinostheo:
Good thinking @cheahenry, let's guard based on permissions in a following PR, for not delaying this one further 😄

PR #225, opened 2022-03-24
Use a wrapper modal for Quantity and Plan Selector based on organization state

dinostheo:
The motive for this PR is to remove the newly introduced dual nature of the Quantity Update modal and have an explicit way to achieve the variation between Quantity Update and Plan Selector from one modal call. 

With the previous fix, under specific organization states, we want to either display the Plan Selector or the Quantity Update from the same action. I previously added this functionality to the Quantity modal. However, I believe that it makes it complicated to explicitly open the quantity modal and I find it myself confusing, calling the quantity modal with configuration on the Publish side for opening the Plan Selector 🤔 .

For that reason I thought moving the wrapper to a new modal called `subscription update` might be easier to reason about..

So if a CTA needs to open the Quantity update for Paying SBB customers and otherwise the Plan Selector, we can use `subscriptionUpdate` modal, otherwise call the specific modals as we currently do.


cheahenry:
So to rephrase we can use either:
- `planSelector`
- `subscriptionUpdate` (that will open either the Plan Selector or the Quantity Update modal)

Correct?

I have an opened [PR](https://github.com/bufferapp/buffer-account/pull/783), I could switch `quantityUpdate` to `subscriptionUpdate`

cheahenry (approved):

PR #227, opened 2022-03-28
Fix/grow 434 refactor logic planselector layouts

daisymarie128:
👀 Because making changes to this logic has been known to cause unforeseen issues for customer would I be able to get some reviews on this PR 😊 

Jira: https://buffer.atlassian.net/browse/GROW-434
Slack threads with more context: https://buffer.slack.com/archives/C0DLE125Q/p1647986280463599

The plan selctor now follows these rules:

If the user is on anything but an agency plan or trial, we show Free | Essentials | Essentials + Team
This should apply to the Plan Selector no matter where it opens from (test with upsell banner, account/billing, upgrade paths, etc.)
The only time the Plan Selector should show the Agency state (Essentials | Essentials + Team | Agency) is if the user is:
1) On an Agency trial
2) On an Agency plan
3) Clicked the CTA in the bottom of the Plan Selector to explicitly view it (“Need more than 10 channels? Try our Agency Plan”)
When a user is on the Free plan only Essentials and Team will be displayed and Essentials will be selected as default

This also contains some new Cypress tests to cover these scenarios

**New plan selector layouts:**
When the user is on an agency plan or trial: 
Screen Shot 2022-03-28 at 3 05 52 pm

When the user is on the free plan and has clicked the Agency CTA in footer:
Screen Shot 2022-03-28 at 3 05 36 pm

When the user is on a Free plan:
Screen Shot 2022-03-28 at 3 05 22 pm

When the user is on the Team plan:
Screen Shot 2022-03-28 at 3 04 44 pm

### DIFF
--- a/packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
+++ b/packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
@@ -174,18 +174,15 @@ export const PlanSelectorContainer = ({
   }, [monthlyBilling]);
 
   useEffect(() => {
-    handleUpgradeIntent(
-      selectedPlan.planId,
-      monthlyBilling,
-      updateSelectedPlan
-    );
-
     updateButton(selectedPlan, channelsCount);
   }, [selectedPlan]);
 
   useEffect(() => {
     updateButton(selectedPlan, channelsCount);
 
+    // When the channel count changes and we are on the Free plan
+    // We do this to automatically switch the user to the essentials plan
+    // as Free plans can have a max of 3 channels
     if (
       selectedPlan.planId === 'free' &&
       channelsCount > selectedPlanMinimumQuantity
--- a/packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
+++ b/packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
@@ -45,9 +45,9 @@ import {
 } from '../../../../../common/utils/user';
 
 import {
-  filterListOfPlans,
   calculateTotalSlotsPrice,
   handleUpgradeIntent,
+  getAvailablePlansForDisplay,
 } from '../../../utils';
 
 export const PlanSelectorContainer = ({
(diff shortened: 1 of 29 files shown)


packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
```diff
@@ -58,26 +57,28 @@ export const PlanSelectorContainer = ({
   trialInfo,
   openSuccess,
   isFreePlan,
-  isUpgradeIntent,
 }) => {
   const { isEnabled: splitSBBEnabled } = useSplitEnabled('slot-based-billing');
   const planOptions = changePlanOptions;
 
+  const shouldIncludeAgencyPlan = isAgencyUser(user) || isOnAgencyTrial(user);
```
dinostheo, line 64: I'm not sure why we make a distinction here. An organization will have an Agency plan subscription if they are paying or trialling for it. I believe checking if they are on the Agency plan is enough for this purpose.

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useInterval.js
```diff
@@ -2,15 +2,10 @@ import { useState } from 'react';
...
-    planOptions,
-    isUpgradeIntent
-  );
+const useInterval = (planOptions) => {
+  const defaultSelectedPlan = getDefaultSelectedPlan(planOptions);
 
-  const initiallyMonthly = isUpgradeIntent
-    ? false
-    : defaultSelectedPlan.planInterval === 'month';
+  const initiallyMonthly = defaultSelectedPlan.planInterval === 'month';
```
dinostheo, line 8: it is a good practice to prefix boolean variables with `is, has, should, ...` etc. Since we are exporting this variable it will make it more obvious that is a boolean when used without the context of this conditional statement.

dinostheo (approved):
I tested this PR and it works exactly as described 🙌 . I've left some minor comments, but they are not blocking. 
My following thoughts are not blocking as well 😄 

So here are m thoughts about this PR:
I did have some difficulties in following the PR since I found that there was a lot happening. 
If I understand correctly what happened here, I would say that
1. The default `interval` is the year interval
2. We display the Agency plan only when the Agency plan is the current plan, or the action to check it on the plan selector is clicked. 
3. When an organization is actively on the Free plan (not cancelling at period end, but already cancelled), display the 2 Essential Plans only.
4. Cleanup the use of the `isUpgradeIntent` 

Simplistically thinking and I might be missing a lot of details, we could approach this work in 3 to 4 different PRs that do not conflict with each other and could be easier to review, test and merge.

packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
cheahenry: Yeah good call, I can simplify that

dinostheo: Something else hit me seeing this line, is that we don't focus if the user or account is on the Agency plan, but on the organization. I believe that might require heavier refactoring for the codebase. 
Our billing though is organization-centric and not user-centric.
A bit pedantic though 😅

PR #236, opened 2022-03-30
[GROW-449] update CTA for tracking

cheahenry:


dinostheo (approved):

cheahenry:
@dinostheo there are quite a few more.
I am going to do an audit and update them in this PR.
After that I will move this PR back to `Ready for review`

dinostheo:
@cheahenry Most of them are tracking a page view event, which is constructing the final CTA for segment. 
e.g. [here](https://github.com/bufferapp/app-shell/blob/d78eb46d329546d8d1e4a51febe020df1aa80497/packages/app-shell/src/components/Modal/modals/StartTrial/index.jsx#L37) that we are passing the properties without a mutation to the server side

Although, I find the approach of having a single CTA, more consistent, changing them all might affect the client side tracking.

PR #237, opened 2022-03-30
Make Plan Selector display on top of navigation

gisete:
We want the Plan Selector modal to always be on top of the navigation.

Related to https://github.com/bufferapp/ui/pull/363


cypress/integration/navigator/navigator.spec.js
```diff
@@ -54,6 +54,7 @@ describe('Navigator', () => {
 
     it('a user on the Free plan should be able click Upgrade and see essentials or team plan displayed', () => {
       cy.contains('Upgrade').should('exist');
+      cy.contains('No Thanks').click();
```
gisete, line 57: @cheahenry @daisymarie128 
I added this here because after removing the z-index rule while the TrialExpired modal was up you can't click on the Upgrade button and the test was failing. Now I make sure to close the TrialExpired modal before clicking on the Navigation button.

PR #241, opened 2022-04-04
Revert counter quantity when selecting away from the Agency plan

gisete:
In this scenario when the user:
changes channel quantity -> selects Agency plan -> selects different plan 
We want to revert the state of the counter to that “edited” quantity

[More context here](https://buffer.slack.com/archives/C0DLE125Q/p1649065125887259)


packages/app-shell/src/common/hooks/useChannelsCounter.js
```diff
@@ -5,14 +5,19 @@ import { getAgencyPlanMinimumChannelInputMessaging } from '../../components/Moda
...
 ) {
   if (planId === 'agency' && channelsCount < 10) {
     setChannelsCounterValue(10);
   }
   if (planId === 'free' && channelsCount > 3) {
     setChannelsCounterValue(3);
   }
+  if (planId !== 'agency') {
+    // change the counter to the last updated quantity number before choosing the agency plan
+    setChannelsCounterValue(lastUpdatedQuantity);
```
daisymarie128, line 19: @gisete We will probably need to add returns in this function as this new code breaks the logic which reverts the channel count to 3 for switching back to the Free plan (see screenshot for what can happen now 😅 ) 
Sorry I haven't written tests for this section of the code but will make a note to add that 😄 

Screen Shot 2022-04-05 at 9 42 29 am

daisymarie128 (changes requested):

PR #242, opened 2022-04-05
[TESTS] Add missing unit tests

daisymarie128:
🚢 

Adding missing unit tests for channelsCounter


dinostheo:
That's great to see 😄 . Thanks @daisymarie128 🌮

PR #246, opened 2022-04-11
[UPDATE] Change button copy to read same as quantity modal

daisymarie128:
Uses sbb FF to simplify logic for this button txt

👀 Would like to get a review on this PR before I merge because I wanted to see if the team would be happy to go with this approach.

I've wrapped this change to the SBB FF. So when we make these changes permanent I was hoping to also clean up some of the logic for this button.

Currently, there is some nested if statements deciding what the text of this button should be. I feel the logic has become quite complex and hard to read. How do we feel about removing all these different button states and just falling back to either `'Confirm and Pay' or 'Go To Payment'` 

Screen Shot 2022-04-11 at 10 49 42 am

Screen Shot 2022-04-11 at 10 49 59 am


dinostheo (approved):
@daisymarie128 thanks for thinking about how to simplify this part 🙏 
Technically, I do agree with this approach. It eliminates a redundant check that we currently make to set the button text as "Stay On My Current Plan" and once again to disable it. 
On top of that on SBB we might stay on the same plan, but just change quantity.

gisete (approved):
Great idea @daisymarie128 🚀

PR #248, opened 2022-04-11
[FIX] Check canEdit before opening planSelector

daisymarie128:
JIRA: [GROW-429](https://buffer.atlassian.net/browse/GROW-429)

How to test:
1. Run appshell locally
2. Visit `https://appshell.local.buffer.com:3000/#planSelector` with a admin user who has a verified email and canManageBilling set to true.
3. The planSelector should open.

Test a user who does not have permission to manage billing:
1. Run appshell locally
2. Visit `https://appshell.local.buffer.com:3000/#planSelector` with a memeber user who does not have a verified email or their canManageBilling is set to false
3. The planSelector should NOT open.

Before:
Screen Shot 2022-04-14 at 10 30 21 am

After:
Screen Shot 2022-04-14 at 10 30 41 am


dinostheo:
@daisymarie128 the core-authentication service provides a new property called `canManageBilling`, which we already query. https://github.com/bufferapp/app-shell/blob/8f3678ff2e29f98bbf749259113964141346b803/packages/app-shell/src/common/graphql/account.js#L108
You can also see how it is being read and used [here](https://github.com/bufferapp/app-shell/blob/8f3678ff2e29f98bbf749259113964141346b803/packages/app-shell/src/components/Modal/modals/PaymentMethod/index.jsx#L58)

This property is set for your account if your account has no admin role (similar to `canEdit`) but additionally when an admin role has not verified their e-mail address. The latter is easier to test as well. You can do that by either signing up with a new account and testing locally (they all have an initially unverified email address) or updating the property `hasVerifiedEmail ` to `false`, in the database for your account. 

Screen Shot 2022-04-11 at 11 43 27

cypress/integration/navigator/teamMembers.spec.js
```diff
@@ -0,0 +1,27 @@
+const REACT_APP_API_GATEWAY_URL_MATCHER = /https:\/\/(graph\.buffer|graph\.local\.buffer).com\//;
+
+describe('Team memebers', () => {
+  describe('NB Team Plan - Team memeber without edit perimissions', () => {
+    before(() => {
+      cy.fixture('accountNBTeamTeamMemeberWithoutEditPermission').then(
```
dinostheo, line 6: 😍

packages/app-shell/src/common/utils/user.js
```diff
@@ -64,3 +66,7 @@ export function getUsersCurrentPlan(user) {
 
   return getSubscriptionPlanData(billingData);
 }
+
+export function hasPermissionToEdit(user) {
```
dinostheo, line 70: This is not a blocking comment. I personally think that the name is not very representative of the privilege. 
```suggestion
export function hasPermissionToManageBilling(user) {
```

or something similar that is more descriptive could be more self-explained when an engineer sees it in use, without its implementation.

dinostheo (approved):
Thanks for adding tests @daisymarie128 🙏

packages/app-shell/src/common/utils/user.js
daisymarie128: @dinostheo I agree with you! good catch! I'll update this quickly then merge 😊

PR #252, opened 2022-04-12
Bump up UI library version

gisete:
- Updated UI library version to v6
- Reverted `z-index` increase in the navigator because the new library version fixes the issue


daisymarie128 (approved):
Thanks for this @gisete 🎉

federicoweber (approved):
Thanks @gisete 🙌

PR #255, opened 2022-04-20
[FIX] Adds wait for cypress request

daisymarie128:
🎪 
This was causing the flakiness in our gh pipeline

Notes:
I managed to reproduce the same flaky failing test locally. I usually always run `yarn test:ux:live` however I found running `yarn run test:ux` produced the same results as running it in the actions pipeline. 
This enabled me to debug the failing test locally and find the root cause which was not waiting for the request to finish before starting each test in the section.

I've also added a separate PR which updates our Cypress docs and adds more tips for debugging errors etc [PR](https://github.com/bufferapp/app-shell/pull/254)


gisete:
Thanks for finding a fix for this @daisymarie128 !! 🏅

PR #256, opened 2022-04-21
Fix/grow 472 hide start subscription

daisymarie128:
👀 
Could I get eyes on this as I made this PR after someone raised the issue in slack. Just wanted to get approval this was the right fix and new user flow.

JIRA: [GROW-472](https://buffer.atlassian.net/browse/GROW-472)
[Slack conversation](https://buffer.slack.com/archives/C0DLE125Q/p1650488171622609)

This fixes the issue where a user could see both the “start subscription” banner and also the verify email banner. Clicking on the “start subscription” cta did nothing and left a user confused.

Includes new Cypress tests for this scenario

The issue:
Screen Shot 2022-04-21 at 10 04 58 am

After fix: 
Screen Shot 2022-04-21 at 10 04 30 am

Original start subscription behaviour still works:
Screen Shot 2022-04-21 at 10 11 57 am

Screen Shot 2022-04-21 at 10 12 03 am


dinostheo:
Hey @daisymarie128 , I mentioned the following in slack but I will post it here as well. 
I believe you spotted well the cause. It is indeed the `canManageBilling`. 
However, I’m thinking that even hiding the start subscription banner might still be confusing for the user since there are multiple CTAs that can open the plan selector or the quantity update modal.

I'm wondering if we should actually open the modals but show a warning message?

We used to do this in the credit card form, but for unverified team members of organizations with a payment method, we were getting the confusing error from the server-side. Would it be better to bring that warning a bit earlier to those modals and explain that they either don't have admin permissions or they have to verify their email?
What do you think about this solution?

cc @cheahenry @gisete

dinostheo (approved):
@daisymarie128 @hamstu 
Code wise this PR looks great.
If we also accept that it does not handle all the CTAs, then we can merge it.

daisymarie128:
> @daisymarie128 @hamstu Code wise this PR looks great. If we also accept that it does not handle all the CTAs, then we can merge it.

Sorry for the late response @dinostheo I lost this PR in a pile of my other ones that came up. 
I agree with your first comments about this is confusing and perhaps we should show a banner to prompt them to show a warning message.

Would we be ok with me closing this PR as it is a bit stale and I think the conflicts in this PR code take about the same time to fix as making the warning message appear and address the issue for all the CTA's?

I will make a jira ticket for this so as to not lose track of it. wdyt?

hamstu:
That sounds good @daisymarie128! 👍🏻

PR #258, opened 2022-04-22
[GROW-468] Install Lodash debounce method

gisete:
This PR aims to install Lodash's debounce method. [GROW-468](https://buffer.atlassian.net/browse/GROW-468)

How to use it:

```
import { useCallback } from "react";
import debounce from 'lodash.debounce';

const handler = useCallback(debounce(debouncedFunctionName, 2000), []);
handler()
```
_I'm making use of `useCallBack` to prevent the debounced function from being called multiple times at every rendering. Passing an empty array as a dependency makes sure the callback is called only once_


daisymarie128 (approved):
Nice one @gisete and thanks for mentioning the `useCallBAck` as well 🙌 looking forward to seeing this get implemented 👍

PR #262, opened 2022-04-27
Task/grow 392 quantity input

daisymarie128:
~[WIP] - looking for input from @gisete before I write cypress tests.~ 

JIRA: [GROW-392](https://buffer.atlassian.net/browse/GROW-392)

This ticket enables users to enter the number of channels they want into the channel input.

Introducing a new context state here. In hopes to make the logic of having to pass props all the way down to children easier.
See the loom video for more context 😊 

Changes in this PR:
[x] Made new PlanInfo context and wrapper PlanSelectorContainer and QuantityUpdate/CardBody with Provider
[x] useChannelsCounter now checks the setChannelsCount and does nothing if an undefined value is received.
[x] Counter now handles input events like onChange and onBlur (for when a user leaves value as empty)
[x] written Cypress tests (todo after getting feedback on new approach)
[x] Written unit test to cover new utils funcs
[x] Refactored useChannelsCounter
      [x] File structure moved
      [x] moved funcs into the utils file - this hook started to get quite busy and felt moving some logic out into the utils file helped make this more readable
  
 
 Might be good if one person could have a quick manual test of this as well (if we have time)
 The reason being, I originally thought this was going to be a straightforward task, however, while implementing the logic I noticed many areas that either didn't fit with existing logic, raised new edge case scenarios, and even though I've written tests to accompany all the changes, this one might benefit from an extra manual test.    
      
Original video - is out of date now, but leaving it here just for some context 😅 
Loom video walking through the change I made and where I was hoping to get some feedback: https://www.loom.com/share/030f9f204b0c46c58261c80622d06104?sharedAppSource=personal_library


packages/app-shell/src/common/hooks/useChannelsCounter.js
```diff
@@ -36,6 +36,12 @@ const useChannelsCounter = (
     useState(null);
 
   function setChannelsCounterValue(value) {
+    // Prevent the channel count from being set to an undefined value
+    // This prevents price calculations returning Nan
+    if (!value) {
```
gisete, line 41: @daisymarie128 do you think it could be helpful to also add other types of validation? For example, we make sure it's a number.

gisete (approved):
This is so great @daisymarie128 !! 🎉  I 100% support us using Context here to share data from the selected plan, it already makes it easier to understand. I have a PR coming that can already benefit from this change so I will wait until you merge it to continue 😄  Thank you!!

packages/app-shell/src/common/hooks/useChannelsCounter.js
daisymarie128: @gisete great thought!! 🙌 

I did some thinking on this while writing it and I think adding js checks to make sure the type is a number would make our code more complex and cumbersome.

In these scenarios where we don't have a type-checking system in place, writing some clear semantic HTML can help us cover this case. For example, here we use the number input which does not allow the user to type anything but a number into it.

And alongside this covering the logic with good unit and UI testing should prevent us from passing back a number.

But cases like this are a good example of how languages like Typescript or a type system can be very beneficial 😊 

Let me know if you think it would be better to add this check though :slightly_smiling_face:

daisymarie128:
> This is so great @daisymarie128 !! 🎉 I 100% support us using Context here to share data from the selected plan, it already makes it easier to understand. I have a PR coming that can already benefit from this change so I will wait until you merge it to continue 😄 Thank you!!

Oh Awesome!! woo thanks for reading over the pr @gisete ! 🌮  I'll finish off the last bit of logic and add some tests 🎉

packages/app-shell/src/common/hooks/useChannelsCounter.js
gisete: Thanks for sharing @daisymarie128, that makes sense. Good call on using `type=number` for the input 🙂

PR #265, opened 2022-04-28
[GROW-388] add assert for counter value for upgrading from free

cheahenry:


cypress/integration/modal/planSelector.spec.js
```diff
@@ -128,6 +128,21 @@ describe('Plan Selector', () => {
 
       // It should default to Essentials year
       cy.get('#essentials_year').should('have.attr', 'aria-label', 'checked');
+
+      // User has only 2 channels but we default to 3 when trying to upgrade
```
dinostheo, line 132: @cheahenry
I know that this is a test for the existing logic, but I believe it is exposing something that doesn't look correct to me. 
Why do we default to 3 channels, for Essentials, when 2 are connected? Especially if we are coming from the Free plan.
Do we perhaps carry the default value set on Free here, wrongly?

dinostheo (approved):
Thanks for adding all these tests (I might need a crash course on using Cypress if you have the time to do some knowledge sharing 😄)

I'm not blocking this PR, since it's a test of existing logic. I do question the current logic though. Am I missing something or understand it wrong?

PR #266, opened 2022-05-03
Show accurate cost when selecting new plan

gisete:


gisete:
@cheahenry I started working with the custom hook you had written but I ended up making so many changes to the original that I decided to create a new hook. I left the one you wrote in case we need it the way it is 🙂  Thanks for coming up with it!

packages/app-shell/src/common/hooks/ChargeAmount/useGetTotalChargeAmount.js
```diff
@@ -0,0 +1,110 @@
...
+            quantity,
+          },
+        });
+      }
+    }
+  }, [selectedPlanId, quantity, hasPlanCounterChanged]);
+
+  useEffect(() => {
+    if (error) {
+      //todo: handle error here
```
daisymarie128, line 98: Not sure if you were leaving this to finish in another PR 😄  but perhaps here we could log a more detailed error to bugsnag i.e :

`      return Bugsnag.notify(
        new Error(
          'usePreviewCustomUpcomingInvoice: getCustomInvoice failed',
          error
        )
      );`

packages/app-shell/src/common/hooks/ChargeAmount/useGetTotalChargeAmount.js
```diff
@@ -0,0 +1,110 @@
...
+
+      return cachedData;
+    }
+
+    return null;
+  }
+
+  useEffect(() => {
+    if (!hasPlanCounterChanged) {
+      //todo skip query when plan/counter remains the same
```
daisymarie128, line 70: Do we still need to keep this todo?

daisymarie128:
Awesome work and thank you for all the refactoring and cleanup @gisete 🙌 🌮 

I think I may have found one bug, but I'm not sure 🤔 

How to reproduce:
1. Be on an account which is on a monthly interval to begin with.
2. Increase your channels. (slowly so you have fetched a few different new total prices)
3. now toggle to yearly plans.
4. Double check the channel quantities you loaded before on monthly. - This is where I think they may be using cached values? 🧐 - but I'm not sure maybe I've just got the pricing wrong 🤔 

Screen Shot 2022-05-12 at 2 35 04 pm

cheahenry:
> @cheahenry I started working with the custom hook you had written but I ended up making so many changes to the original that I decided to create a new hook. I left the one you wrote in case we need it the way it is 🙂 Thanks for coming up with it!

No worries, thanks for letting me know. It's not being used so feel free to delete it!

gisete:
> Awesome work and thank you for all the refactoring and cleanup @gisete 🙌 🌮
> 
> I think I may have found one bug, but I'm not sure 🤔
> 
> How to reproduce:
> 
> 1. Be on an account which is on a monthly interval to begin with.
> 2. Increase your channels. (slowly so you have fetched a few different new total prices)
> 3. now toggle to yearly plans.
> 4. Double check the channel quantities you loaded before on monthly. - This is where I think they may be using cached values? 🧐 - but I'm not sure maybe I've just got the pricing wrong 🤔
> 

Thank you @daisymarie128 for finding this bug! I thought this had to do with the fact that I hadn't include the plan cycle as an argument in `useEffect` so I fixed that but it turns out this was related to a more serious bug. [More context here](https://buffer.slack.com/archives/C0DLE125Q/p1652399529153399)

PR #268, opened 2022-05-04
[GROW-484] Cleanup the SBB usage and make it the default behaviour

dinostheo:
Removes the usage of the SBB feature flip and sets the SBB behaviour as the default one. 
[[GROW-484](https://buffer.atlassian.net/browse/GROW-484)]

NOTE:
Have a thorough look at the changes and give it a try locally to see if everything works as it should.


cheahenry (approved):
Nice, so much cleaning!
Tested locally :)

PR #269, opened 2022-05-04
[GROW-493] Clean the querying of the changePlanOptions.channelsQuantity

dinostheo:
Stops querying the `changePlanOptions.channelsQuantity`. 
This is one part of deprecating this property.


cheahenry (approved):
Thanks for doing this!
Let's make sure we find a better solution instead of doing it manually.
It can wait until we move the relevant pieces into the mono-repo.

PR #270, opened 2022-05-05
Task/error handling large channel quantity

daisymarie128:
[Slack conversdation](https://buffer.slack.com/archives/C0DLE125Q/p1651671143739739?thread_ts=1651644082.105319&cid=C0DLE125Q)

This PR adds an error message and disables the submit button when a user tries to  enter a value in the channel input, greater than or equal to `8333` which is the maximum amount of channels which we are able to process through Stripe

Screen Shot 2022-05-05 at 3 32 25 pm


cypress/integration/modal/planSelector.spec.js
```diff
@@ -348,13 +348,9 @@ describe('Plan Selector', () => {
     it('should display warning message when trying to decrease quantity below 10', () => {
       cy.get('#render_plan_selector').click();
 
-      cy.get(
-        '.UXMessagingstyle__UXMessagingContainer-app-shell__sc-nrg3jr-0'
-      ).should('not.exist');
+      cy.get('#UXMessagingContainer').should('not.exist');
```
cheahenry, line 351: Nice, thanks for adding an id. Its way cleaner!

cheahenry (approved):
LGTM!

Tested locally for the plan selector and the quantity update modal.
Thanks for adding a Cypress test :)

dinostheo (approved):
LGTM @daisymarie128 🌮

PR #273, opened 2022-05-09
Task/grow 480 install typescript

daisymarie128:
This PR aims to add Typescript to the app-shell repo.

👀 Would love some eyes on this to check if I've added all the correct settings/rules for starting to add Typscript to the repo.

In this PR:
[x] Adds Typescript to root of repo
[x] Adds Typescript to packages/app-shell
[x] Adds Typescript to packages/app
[x] Adds eslint rules
[x] Converts one file: common/utils/validation into .ts file and adds types


packages/app-shell/tsconfig.json
```diff
@@ -0,0 +1,107 @@
+{
```
cheahenry, line 1: What tsconfig did you use? 
@federicoweber do we have a default one we want to use for all of our frontends?

cheahenry:
Whenever you merge this, it would be good to squash at the same time. I don't think all the commit history is needed.
Screenshot 2022-05-18 at 09 09 12

packages/app-shell/tsconfig.json
daisymarie128: I used the auto generated one and left all the comment descriptions in just for the moment incase we needed to add or change anything. But I followed most the same rules as what Fed had in buffer-publish: https://github.com/bufferapp/buffer-publish/pull/2221/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

federicoweber: Good question @cheahenry.
The config in `buffer-standards`, is the ideal state, and it's what we are following in Start Pages, but we are having some troubles with that in Publish, specifically around `strict`.
If for some reason we can't get strict to work the recommended fallback is to enable
```
    "strictFunctionTypes": true,
    "strictNullChecks": true,
```

packages/app-shell/tsconfig.json
```diff
@@ -0,0 +1,107 @@
...
+  // TODO: Clean up this file after we are happy with the settings
```
federicoweber, line 2: Hey @daisymarie128 would you mind cleaning up all the commented lines before merging this in?

.github/workflows/cypress.yml
```diff
@@ -5,14 +5,33 @@ on: [push]
 jobs:
   cypress:
     runs-on: ubuntu-latest
-    container: cypress/browsers:node12.13.0-chrome78-ff70
+    container: cypress/browsers:node14.17.0-chrome88-ff89
```
daisymarie128, line 8: I originally tried to use the latest version `cypress/browsers:node16.5.0-chrome97-ff96`

But I found using this version doesn't provide any helpful info when the step fails. Which makes it very hard to debug and wastes a lot of time. (see two images for example)

For now, I've used a lower version here just because it shows the lines which are causing that step to break.


Images show the same failing step but run with the different versions and what errors they print to the console.

I will add a comment to our docs (unless a better place for it) just noting changing this version may cause this issue again.

Screen Shot 2022-05-24 at 2 48 53 pm

Screen Shot 2022-05-24 at 2 48 42 pm

.github/workflows/cypress.yml
```diff
@@ -5,14 +5,33 @@ on: [push]
...
+
     steps:
-      - name: Host
-        run: apt update && apt install sudo && sudo echo "0.0.0.0 publish.local.buffer.com" >> /etc/hosts && sudo echo "0.0.0.0 analyze.local.buffer.com" >> /etc/hosts && sudo echo "0.0.0.0 engage.local.buffer.com" >> /etc/hosts
+      - name: Update packages and install sudo
+        run: |
+          apt update
+          apt install sudo
+
+      - name: Add hosts to /etc/hosts
```
daisymarie128, line 17: This code isn't related to installing typescript. But because of some funny errors I was getting I ended up tidying these steps up a little bit to help me debug

types/modules.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@bufferapp/ui/style/colors';
```
daisymarie128, line 1: Has to add this file to fix this error about a file declaration for this package.
Screen Shot 2022-05-24 at 3 12 16 pm

types/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+declare global {
+    interface Window {
+        appshell: Appshell;
+    }
+
+    interface WindowEventMap {
+        'appshell__organization_event': CustomEvent;
```
daisymarie128, line 73: @federicoweber Do you know if a package outside of this repo, that uses the appshell. (publish for example) 

If I have missed a key type definition here for the `window.appshell` will that break their builds? 🧐 Or do these changes have no risk of breaking applications which import this packages?

.eslintrc.js
```diff
@@ -29,12 +30,31 @@ module.exports = {
...
+    'no-console': ['error', { allow: ['warn', 'error'] }],
   },
   ignorePatterns: ['lib/*'], // Stop ESLint complaining when looking at transpiled lib
   settings: {
     'import/resolver': {
-      webpack: {
-        config: './packages/app-shell/webpack.config.common',
+      // webpack: {
+      //   config: './packages/app-shell/webpack.config.common',
+      // },
```
cheahenry, line 54: Is this still needed?

cheahenry (approved):
LGTM, great job!

types/index.d.ts
federicoweber: That's for adding the types @daisymarie128 🙌.
It look's like this has all we need, so far I was declaring a [global override](https://github.com/bufferapp/buffer-publish/blob/dc418ca21fab54489bc006460060b633cbbfb5c1/packages/web/components/App/index.tsx#L18-L22) with type `any` when using the app shell with typescript.

It shouldn't break compilation, as we are not currently importing the `appshell` as a package, but directly injecting it from a static .js file 🙂. 

It could be worth start publishing those types under a dedicated package `@bufferapp/type/appshell`, so that we could import the packages in the rest of the apps, until we migrate it all to the monorepo 🙂.

The only thing I will suggest is to collocate the types closer to where we are using them for example we could move `interface Modals` to [useModals](https://github.com/bufferapp/app-shell/blob/main/packages/app-shell/src/common/hooks/useModal.js), but export those types form here as well.

types/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+    planSelector: 'planSelector',
+    quantityUpdate: 'quantityUpdate',
+    startTrial: 'startTrial',
+    success: 'success',
+    trialExpired: 'trialExpired',
+    upgradeSuccess: 'upgradeSuccess',
+    // This modal will make a decision based on the Organization state if
+    // it will render the planSelector or the QuantityUpdate
+    subscriptionUpdate: 'subscriptionUpdate'
+}
```
federicoweber, line 20: I think this could be a good use case for [enums](https://www.typescriptlang.org/docs/handbook/enums.html) 😉

types/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+
+interface ActionOptions {
+    cta?: string,
+    ctaButton?: string,
+    shouldPickModalOnOrganizationState?: boolean,
+}
+ 
+interface RenderComponentParams {
+    componentKey?: string,
+    containerId?: string,
```
federicoweber, line 34: those two should be mandatory 🙂

types/index.d.ts
federicoweber, line 20:
```suggestion
enum Modals {
    GEID1FreeTrialPrompt ='GEID1FreeTrialPrompt',
    channelConnectionPrompt = 'channelConnectionPrompt',
    essentialsPlan = 'essentialsPlan',
    essentialsPricing ='essentialsPricing',
    paidMigration = 'paidMigration',
    paymentMethod = 'paymentMethod',
    paywall = 'paywall',
    planSelector = 'planSelector',
    quantityUpdate = 'quantityUpdate',
    startTrial = 'startTrial',
    success = 'success',
    trialExpired ='trialExpired',
    upgradeSuccess ='upgradeSuccess',
    // This modal will make a decision based on the Organization state if
    // it will render the planSelector or the QuantityUpdate
    subscriptionUpdate ='subscriptionUpdate'
}
```

types/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+    trialExpired: 'trialExpired',
+    upgradeSuccess: 'upgradeSuccess',
+    // This modal will make a decision based on the Organization state if
+    // it will render the planSelector or the QuantityUpdate
+    subscriptionUpdate: 'subscriptionUpdate'
+}
+    
+interface Components {
+    startTrialButton: 'startTrialButton',
+}
```
federicoweber, line 24:
```suggestion
enum Components {
    startTrialButton = 'startTrialButton',
}
```

types/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+    componentKey?: string,
+    containerId?: string,
+    options?: ActionOptions,
+}
+    
+interface AppshellEvents {
+    MODAL_EVENT_KEY: 'appshell__modal_event',
+    ORCHESTRATOR_EVENT_KEY: 'appshell__modal_event',
+    ORGANIZATION_EVENT_KEY: 'appshell__organization_event',
+}
```
federicoweber, line 42:
```suggestion
enum AppshellEvents {
    MODAL_EVENT_KEY = 'appshell__modal_event',
    ORCHESTRATOR_EVENT_KEY = 'appshell__modal_event',
    ORGANIZATION_EVENT_KEY = 'appshell__organization_event',
}
```

types/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+    openModal: (modalKey: string, modalData?: ActionOptions) => void,
+    renderComponent: (data: RenderComponentParams) => void,
+    setCurrentOrganization: (organizationId: string, options?: ActionOptions) => void,
+    billingUpdated: (options?: ActionOptions) => void,
+}
+    
+interface AppshellActionKeys {
+    setCurrentOrganization: 'setCurrentOrganization',
+    billingUpdated: 'billingUpdated',
+}
```
federicoweber, line 54:
```suggestion
enum AppshellActionKeys {
    setCurrentOrganization = 'setCurrentOrganization',
    billingUpdated = 'billingUpdated',
}
```

federicoweber (approved):
Great work @daisymarie128, I left some small comments inline that I would love for us to address before merging, but overall this looks great, thank you 🙌.

.eslintrc.js
daisymarie128: Oh thanks for raising this @cheahenry 🙌 
I forgot to leave a note here about this. I had to remove this to fix for ts linting to work in `packages/app`. Removing it didn't seem to break anything but It was added 3 months ago and I wasn't sure of its actual purpose 🤔

types/index.d.ts
```diff
@@ -0,0 +1,76 @@
...
+declare global {
+  interface Window {
+    appshell: typeof Appshell;
+  }
+
+  interface WindowEventMap {
+    // eslint-disable-next-line camelcase
+    appshell__organization_event: CustomEvent;
+  }
+}
```
hamstu, line 76: I think you only need `typeof` when it's before an enum ([ref](https://www.digitalocean.com/community/tutorials/how-to-use-enums-in-typescript#extracting-the-object-type-of-enums)). 

So not needed in front of `Appshell` on line 69, or  `AppshellActions`

hamstu:
I think we can close as this is being worked on elsewhere now. cc @jacobchadwell for context.

PR #274, opened 2022-05-10
[GROW-513] add prefixWithProduct function

cheahenry:
We want to prefix with the name of the product on the client and server side.


dinostheo (approved):
LGTM @cheahenry

PR #277, opened 2022-05-11
Refactoring use lazy query

daisymarie128:
It's not super polished but the way I tried to work with this was to implement a cache lookup, which then prevents us from firing the denounced query and triggering any loading states.
Let me know if you have any questions at all 😊 

loom video walking through some of the changes: https://www.loom.com/share/2b4e022e66684bb3a11407dc529b2c51


gisete:
Thank you @daisymarie128 ! I feel like this should have been easier to accomplish other than having to do a cache lookup but I like this solution and it works! 🙂  
Also liked the suggestion of following the util functions pattern, easier to do unit testing this way.

PR #283, opened 2022-05-11
[COREP-822] Add TikTok logo to Channel Connection Prompt on publishing calendar

mickmahady:
**Jira Ticket**
https://buffer.atlassian.net/browse/COREP-822


mickmahady:
Closing until we're ready to launch so we can free up some deployment space for other test branches

PR #285, opened 2022-05-12
[GROW-503] Re-open paywall if the planselector is opened by it without conversion

dinostheo:
It fixes an issue where, when the plan selector was dismissed after being opened by a paywall action, it would give access to the paywalled product. 
The solution is to pass an extra property (`source`) to the modal when it's opened via the Paywall's Plan Selector CTA. When this property is passed and the Plan Selector is closed or dismissed then the 

[[GROW-503](https://buffer.atlassian.net/browse/GROW-503)]

https://www.loom.com/share/66ce3e47a8244f6ab49a2bd8589b67ff


cheahenry (approved):
Nice, I like the solution you came up with!
Could you add a test for this if its not too complicated please?

dinostheo:
@bufferapp/labs-growth 
This approach works, but the way we have modified the code here feels very hacky, just to write tests. 
It feels that this specific test would have been better performed in the Analytics or Engagement codebases where the application is actually running or in a test environment, where the complete Buffer app is running. 

I will close this PR and go back to the initial commit, in a new PR.

PR #288, opened 2022-05-18
[GROW-509] Display the Invite Team CTA for Essentials with an upgrade path action

dinostheo:
It displays the `Invite team members` CTA for the Essentials plan as well. However, instead of redirecting the user to the `Team Manage` page, it opens the Plan Selector in order to upgrade. 

[[GROW-509](https://buffer.atlassian.net/browse/GROW-509)]

After this is merged, it will only start being effective when the server side changes on the field have been deployed as well. 
https://github.com/bufferapp/core-authentication-service/pull/1673


cheahenry (approved):
LGTM, I just find it a bit difficult to follow because there is product knowledge/context to have, I had to check the implementation of these to know what it does or example:
- `canHaveTeamMembers`
- `currentOrganization.shouldDisplayInviteCTA`

dinostheo:
@cheahenry I do agree. I didn't even remember that we had `shouldDisplayInviteCTA ` in place and I initially implemented it. 
We discussed about it yesterday in the WG, that we often reason the product behaviour by looking into the code, instead of the other way around. 

Finally, [I created a task](https://buffer.atlassian.net/browse/GROW-530) to move the `canHaveTeamMembers` on the server side. We should reduce this type of logic from the client side where possible, as it makes it difficult to maintain.

PR #292, opened 2022-05-19
[GROW-532] add upgradePathViewed event for trial expired modal

cheahenry:
Screenshot 2022-05-19 at 12 11 10


packages/app-shell/src/components/Modal/modals/TrialExpired/index.jsx
```diff
@@ -80,6 +80,13 @@ export const Modal = ({ user, onDismiss, onUpgrade }) => {
       },
       user,
     });
+    useUpgradePathViewed({
+      payload: {
+        name: 'trial expired',
+        upgradePathName: 'navBar-trialExpiredModal',
```
dinostheo, line 86: @cheahenry does the pathName has different format than a CTA?

dinostheo (approved):
LGTM @cheahenry 
I have a question about the pathname format, as it looks like it that is missing some information in comparison to the CTAs.

packages/app-shell/src/components/Modal/modals/TrialExpired/index.jsx
cheahenry: Good question! `upgradePathName` doesn't have a format (yet). I wrote down my thoughts [here](https://buffer.atlassian.net/browse/GROW-532?focusedCommentId=43061).

Feel free to chime in

PR #295, opened 2022-05-31
[WIP] integrate hermes

erickhun:
cc @cheahenry


hamstu:
Do you need help with that error @erickhun? The server looks for the HTTPS cert that's in the reverseproxy dir (as seen in the log output), but maybe your local setup with hermes has a diff file structure? The reason it needs the cert is so that you can load appshell.local.buffer.com without warnings.

PR #297, opened 2022-05-31
[GROW-539] UX improvement to the accurate cost calculation 

gisete:
Couple of improvements on how we're displaying the final plan cost in the plan selector.

- Polished the final cost style, increased font size and weight
- added a tooltip to display a note about the total cost including discounts and credits.

image


cheahenry (approved):
LGTM! Let’s do some investigation on why cypress failed in the first place. It seems really odd to me.

PR #300, opened 2022-06-15
Update Channels.jsx

daisymarie128:
Add tiktok logo to modals
We want the Tiktok logos in the modals ready for the launch next week 🙂

👀 I think this is all that needs to be done for grow-559 - however, I have not visually checked everything looks correct.

Criteria to merge:
[ ] Plan selector modal displays the Tiktok logo (layout should not be broken)
[ ] Channel Quantity modal displays the Tiktok logo (layout should not be broken)


cheahenry (approved):
Screenshot 2022-06-15 at 08 23 02
(this needs a bit a left margin)
Screenshot 2022-06-15 at 08 23 12

PR #319, opened 2022-06-23
[SP 665] appending ids to Start Page tab on Nav Bar

gmzjuliana:
We're attaching some pendo ids to communicate to users the upcoming changes to Start Page as a channel 😊 I added a new component as the logic is very particular and this won't live forever, likely a couple weeks. Let me know if anything is unclear!


ssvmvss:
Heya @gmzjuliana! 👋 Great call about isolating this code since as you've said it's going to be removed in a month approx! 🙂 

By checking this though I'm wondering if isolating the code on a hook would make more sense? 🤔 I'm thinking that because we are only adding some logic here, we don't have a need for changing the rendering of the Start Page link itself. What if we do something like this instead?:

```
QUERY_START_PAGES = gql`...`

function usePendoModals () {
   const [startPages, setStartPages] = useState();

   useEffect(() => {
      ... Query start pages ...
   }, []);
   
   const getIdForPendo = () => {
      ... Check start pages and do calculations ...
   }
   
   return {
      getIdForPendo
   }
}
``` 

And then, on any component that renders in the AppShell (these Ids do not need to be added to the Start Page link, we could even use the main wrapper! 🙂 ):

```
<Component id={getIdForPendo()} />
```

This is a suggestion by the way! 🙂 If you still prefer to keep this as a component, the only change that I would ask for would be to rename it so it's clearer what the component does. Since in the end the component is just rendering the Start Page link maybe we can just rename it to `StartPageLink`? What do you think?

ssvmvss (changes requested):
Thanks so much for working on this @gmzjuliana! 🙌 🌮 Since it's super important to make sure we add the correct ids for the different types of users (we want to make sure we get the communication right! 😅) I'm wondering if you would be up for adding tests? One per each group at least? 🙏

gmzjuliana:
hey @ssvmvss thanks for the nudge on the tests! I saw they only use Jest and so I extracted the function and am testing that. Let me know what you think :)

I also left it as a component for now but if we start to use this process to communicate through pendo it'd definitely be good to keep in mind!

packages/app-shell/src/components/NavBar/components/NavBarProducts/StartPageLink.jsx
```diff
@@ -0,0 +1,91 @@
...
+        id = '';
+        break;
+    }
+  });
+  return id;
+};
+
+export const StartPageLink = () => {
+  // eslint-disable-next-line no-underscore-dangle
+  const orgId = window.__userData.data.account.currentOrganization.id;
```
dinostheo, line 75: I believe we could use the `useUser()` method for the user context here. Then extract the organizationId from it.

e.g. something like the following
```
const user = useUser()
const orgId = user?.currentOrganization.id
```

dinostheo (approved):
Hey, @gmzjuliana thanks for working on this. It LGTM. 

I have two non-blocking comments, with the most important to my opinion 😅 , to be the one about using the user context.

packages/app-shell/src/components/NavBar/components/NavBarProducts/StartPageLink.jsx
gmzjuliana: oooh we pressed send/merge at the same time almost! sorry @dinostheo, but thanks for the heads up, i'll write a small follow up cause I too like that better 😁 I forgot about it!

packages/app-shell/src/components/NavBar/index.test.js
```diff
@@ -65,3 +66,75 @@ describe('NavBar', () => {
     });
   });
 });
+
+describe('getPendoIds', () => {
+  test('returns noImpact for users who were not at their plan limit', () => {
```
dinostheo, line 71: @gmzjuliana another comment I thought I sent, but apparently, I didn't was about the mixed use of `test` and `it` here. Would be good to have a consistent code style at least per file if we are moving to the use of a new style (e.g. if we are going to use `test` instead of `it` in our tests)

gmzjuliana: oooh good catch! thanks @dinostheo

PR #337, opened 2022-06-28
[GROW-580] Update plan selector copy for yearly plans

jacobchadwell:
Add "(charged yearly)" copy to the end of the priceNote for yearly plans.


packages/app-shell/src/components/Modal/modals/PlanSelector/components/SelectionScreen.jsx
```diff
@@ -116,7 +116,7 @@ export const SelectionScreen = ({
             highlights={option.highlights}
             currency={option.currency}
             basePrice={option.basePrice}
-            priceNote={option.priceNote}
+            priceNote={monthlyBilling ? option.priceNote : `${option.priceNote}\n(charged yearly)`}
```
cheahenry, line 119: All `option` data is coming from the [backend](https://github.com/bufferapp/core-authentication-service/blob/5ed2d4789395fab05bd7b3b4b55a4698f7281270/core-api/src/graphql/schema.graphql#L406). I feel like it would be better to make the change there so it's in one place. What do you think?

jacobchadwell: That makes sense to me! I'll update this PR to only include the style change I made, then work with you on that backend change

cheahenry (approved):
Awesome, thanks!
For PR reviews, feel free to tag the team `@bufferapp/labs-growth`

PR #347, opened 2022-06-30
add logic to append an id to start page tab

gmzjuliana:
[As raised in this Slack conversation](https://buffer.slack.com/archives/C0YS0TW5P/p1656575071388829), we'll need to append an id to the Start Page tab so we can show free users at limit a pendo modal communicating the upcoming changes. 

https://buffer.atlassian.net/browse/SP-690


gmzjuliana:
would someone be up for helping me with the cypress tests? Is it okay that they don't pass? @dinostheo @daisymarie128

gmzjuliana:
hi again! so the first errors were legit but this one, I don't know 😬 I asked the infra team here but again, any context is helpful! https://buffer.slack.com/archives/C0H8K9H8F/p1656624156021249

packages/app-shell/src/components/NavBar/components/NavBarProducts/StartPageLink.jsx
```diff
@@ -71,6 +71,25 @@ export const getPendoId = (startPages) => {
...
 
+// a week before the launch, we want to let free users at their
+// limits know what's going to happen to their SP via a message in Pendo
+// i also added a task in the release plan to remove this at launch so they can get
+// the correct pendo modal then
+// https://paper.dropbox.com/doc/Start-Page-As-A-Channel-Implementation-plan-No-Legacy-ing--BkbQL_KcAzURbWLv23R5Qv_RAg-By8T1RKJhQttRHRneO3pW#:uid=059762319064679039822417&h2=3)-Public-release
+export const getFreeOrgWarning = ({ user }) => {
+  const isOneBuffer = user?.currentOrganization?.isOneBufferOrganization;
+
+  const isFree = isOneBuffer
```
dinostheo, line 82: We do have a method ([isFreeUser](https://github.com/bufferapp/app-shell/blob/main/packages/app-shell/src/common/utils/user.js#L3)) for New Buffer customers to check their plan and not having to access the data structure directly. 
We don't have that for MP customers, but we could simply create one. I think that would be useful even if that code is not here to stay, since it promotes to accessing the data structure directly as little as possible.

packages/app-shell/src/components/NavBar/components/NavBarProducts/StartPageLink.jsx
```diff
@@ -71,6 +71,25 @@ export const getPendoId = (startPages) => {
...
+    ? user?.currentOrganization?.billing?.subscription?.plan?.id === 'free'
+    : user?.currentOrganization?.billing?.subscriptions[0]?.plan === 'free';
+  const isAtLimit = user?.currentOrganization?.channels?.length >= 3;
```
dinostheo, line 85: Similarly, there is a method to access the [channelsCount](https://github.com/bufferapp/app-shell/blob/main/packages/app-shell/src/common/utils/user.js#L100) and having to access the data structure directly. 
Also although we are within the scope of the Free plan limit, it might be more explicit to make that evident via the variable name, e.g. `isAtFreePlanChannelLimit` or something similar. I'm not very good at naming myself 😅 , but I'm missing some information about where this `3` is coming from here.

dinostheo (changes requested):
This code works great @gmzjuliana and thanks for adding all these tests 🌮 

I made some comments about accessing the data structure directly. I know it's difficult to know about them, but there are some access methods in place which you can use to fetch the same information.

gmzjuliana:
thanks for letting me know about those @dinostheo ! I've made the changes requested but I'll test and merge in case you can't approve quickly 😊

dinostheo:
Thanks so much for working on it @gmzjuliana 🙌

PR #351, opened 2022-07-05
move id to own span

gmzjuliana:
the pendo guide isn't currently working because you can only have 1 id per element and it can't be space separated like classes


dinostheo (approved):
lgtm @gmzjuliana

PR #358, opened 2022-07-07
[GROW-611] Show upgrade paths to free mobile customers

gisete:
Show different experience only for paid `revenueCat` customers

This is wrapped in the `cross-platform-upgrades` split


gisete:
@daisymarie128 I left you a comment in Jira but this is the PR for it. I did a search in the code and this component is the only one using `revenueCat` to display a different experience. Please let me know if you had something else in mind 🙂

daisymarie128:
@gisete Nice one! Thanks for adding this in. I think that should be the only one we need to address for this ticket. I think we could merge this one. Though I'm having some local dev issues atm so haven't been able to test this one locally yet 👍

daisymarie128 (approved):

PR #369, opened 2022-07-20
show My Preferences on all tabs

jacobchadwell:
Pending JIRA ticket to link


hamstu:
Hey @jacobchadwell - curious about this one! I'm not sure this will work right outside of Publishing, but maybe I'm missing some context. (Perhaps we want it to always show and go to Publish?)

PR #370, opened 2022-07-21
[GROW-643] Fix - Disable agency link when paying on mobile

jacobchadwell:
https://buffer.atlassian.net/browse/GROW-643


hamstu (approved):
Thanks for the quick fix @jacobchadwell! 💯

PR #379, opened 2022-08-03
Feature/grow 638 refetch account on plan selector

jacobchadwell:
Cause app-shell to refetch user's account data when the Plan Selector modal is opened.

**Purpose:**
When a user adds a credit card in the Account -> Billing page and immediately goes to the Plan Selector, the Plan Selector shows the credit card input field. By refetching the account data before rendering the modal, we make sure the Plan Selector uses the user's saved credit card.


packages/app-shell/src/common/hooks/useModal.js
```diff
@@ -30,6 +30,15 @@ function useModal({ refetchAccountData }) {
   const openModal = useCallback((modalKey, modalData = null) => {
     const matchingModal = Object.keys(MODALS).find((k) => k === modalKey);
     if (matchingModal) {
+
+      // Refetch account data on planSelector open to make sure we have the latest data,
+      // particularly the user's credit card if they just added it
+      // Note: this is only for modals opened with the openModal action, not for event-opened
```
jacobchadwell, line 36: (I did check and we only ever open the Plan Selector modal using this openModal action vs. the event dispatch method)

PR #386, opened 2022-08-16
Task/grow 663 target location

daisymarie128:
JIRA: [GROW-663](https://buffer.atlassian.net/browse/GROW-663)

This PR adds a new util to check whether a customer is based in India. This is for Payment Laws relating to RBI work.

We use the users local time/date settings to grab their timezone and then check if it is from India. This method isn't 100% accurate but might be sufficient WDYT?

Decided to go with just adding Indian timezones instead of functionality for all timezones, as we eventually may not need to target any countries later on.

**How to test**
Change your timezone settings in your system settings
Screen Shot 2022-08-16 at 3 36 44 pm

You can type `Intl.DateTimeFormat().resolvedOptions().timeZone` into your browser console to see which zones you're in.

**Also checked Timezones in Mixpanel Match Indian customers**
Screen Shot 2022-08-16 at 6 12 30 pm


cheahenry (approved):
LGTM,

During our sync, we checked that all users with country code `IN` in [MixPanel are in this timezone](https://mixpanel.com/project/1959816/view/37751/app/insights#~(columnWidths~(bar~())~description~'~displayOptions~(chartType~'bar~plotStyle~'standard~analysis~'linear~value~'absolute)~sorting~(bar~(sortBy~'column~colSortAttrs~(~(sortBy~'value~sortOrder~'desc)))~line~(sortBy~'value~sortOrder~'desc~valueField~'averageValue~colSortAttrs~(~))~table~(sortBy~'column~colSortAttrs~(~(sortBy~'label~sortOrder~'asc)))~insights-metric~(sortBy~'value~sortOrder~'desc~valueField~'totalValue~colSortAttrs~(~))~pie~(sortBy~'value~sortOrder~'desc~valueField~'totalValue~colSortAttrs~(~)))~timeComparison~null~querySamplingEnabled~false~title~'~sections~(show~(~(dataset~'!mixpanel~value~(name~'!all_people~resourceType~'people)~resourceType~'people~profileType~'people~search~'~dataGroupId~null~math~'total~perUserAggregation~null~property~null))~group~(~(dataset~'!mixpanel~value~'!timezone~resourceType~'people~profileType~null~search~'~dataGroupId~null~propertyType~'string~typeCast~null~unit~null))~filter~(clauses~(~(dataset~'!mixpanel~value~'!country_code~resourceType~'people~profileType~null~search~'~dataGroupId~null~filterType~'string~defaultType~'string~filterOperator~'equals~filterValue~(~'India)~propertyObjectKey~null))~determiner~'all)~time~(~(value~30~unit~'day)))~legend~()))

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
```diff
@@ -21,7 +24,8 @@ const useButtonOptions = ({
     getLabel(selectedPlan, currentChannelQuantity, updatedChannelQuantity)
   );
 
-  const paymentType = isPaymentCheckoutEnabled ? checkoutSession : openPaymentMethod;
+  const shouldUsePaymentCheckout = isPaymentCheckoutEnabled && isBasedInIndia() && isFreeUser;
```
cheahenry, line 27: You can add a comment that the free check can be removed when when we will do the work on the backend to cancel the previous subscription

PR #387, opened 2022-08-18
[UPDATE] Refactor to use attributes in Split

daisymarie128:
This PR refactors how we are checking if a user should see the new checkout experience.
Relates to previous GROW-663 PR that was merged.

This PR adds the `isInIndia()` boolean to our useFeatureEnabled hook and allows us to add the logic inside of Split.io dashboard instead.
Values were also added to Split.ios targeted traffiking 

@hamstu Would you be able to give this a quick look over to make sure I've got the syntax correct in the code but also on Split.io? 👀 
I only added the targets on local for now as I wasn't sure if I add it to Prod now if it will start serving customers the FF right away? 🤔 

Screen Shot 2022-08-18 at 2 31 43 pm


packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/useButtonOptions.js
```diff
@@ -15,19 +15,15 @@ const useButtonOptions = ({
   currentChannelQuantity,
   updatedChannelQuantity,
   checkoutSession,
-  isFreeUser,
 }) => {
-  const { isEnabled: isPaymentCheckoutEnabled } = useSplitEnabled('payment-checkout')
+  const { isEnabled: isPaymentCheckoutEnabled } = useSplitEnabled('payment-checkout', { isIndianTimezone: isBasedInIndia() })
```
cheahenry, line 19: I think you also have to send `isFree` if you want to use it in Split, otherwise it won't have the data.

hamstu:
LGTM - looks great @daisymarie128!

Quick tip for [testing from Split's side too](https://www.loom.com/share/e6ac91da503f47da8cab3513dc0ad0a2) (Loom - 2min).

PR #421, opened 2022-09-20
[UPDATE] Fix tracking utils func checks

daisymarie128:
👀 Would love a second set of eyes on this PR. Just as if I merge I'm not to sure how to get an MP account in the correct state to test that everything is still working.
If someone could just look over the code and check none of the changes look like they would cause tracking to break 👍 

This PR aims to fix parts of our tracking which may be returning incorrect data for users. In particular, this PR looks at the util functions we are using to grab information about trials.

**More context about the code changes:**
I noticed the original util was using this path to find if a user was on a trial - `currentOrganization?.billing?.subscription?.plan?.trial?.isActive` which is **NOT** correct for NEW BUFFER users.

However, it is correct for MP Users and similar to the change made for this function I also had to ammend the way we find the interval as well.

**MP Users data structure**
Screen Shot 2022-09-20 at 2 08 12 pm

**New Buffer data structure**
Screen Shot 2022-09-20 at 2 20 51 pm


cheahenry (changes requested):
Good find! However, I don't think your fix will work. 
Here's how we query the data:
https://github.com/bufferapp/app-shell/blob/da47217c022346fb83c8c957061095a7b62c14b9/packages/app-shell/src/common/graphql/account.js#L44-L63

It looks like for MP, the data is in subscription**s**.interval and subscription**s**.trial.isActive
For NB, it's in subscription.interval and subscription.trial.isActive

We can also verify how the data is in the API, here's the subscriptions types:
https://github.com/bufferapp/core-authentication-service/blob/e2ad31140532ca7737f9db8219daee9bc03f9d38/core-api/src/graphql/schema.graphql#L550-L578

It's subscription**s** for MPBilling: 
https://github.com/bufferapp/core-authentication-service/blob/e2ad31140532ca7737f9db8219daee9bc03f9d38/core-api/src/graphql/schema.graphql#L358

And subscription for OBBilling: https://github.com/bufferapp/core-authentication-service/blob/e2ad31140532ca7737f9db8219daee9bc03f9d38/core-api/src/graphql/schema.graphql#L328

Let me know if you have any questions!

daisymarie128:
@cheahenry Thanks for reviewing this and catching that mismatching path 🙌  and sharing all those type definitions, super helpful 🌮

hamstu:
Great work here @daisymarie128 and thank you @cheahenry for that extra context on the expected data. If this is feeling good then let's merge it in first thing.

As part of the data audit I'm doing with Bats/Julian knowing that this will be accurate is really helpful. We may have some more tasks like this coming up, and it'll be helpful to think about the differences between MP/NB :)

PR #423, opened 2022-09-21
[GROW-709] Make migration modal more generic

cheahenry:
The migration modal was originally developed only to support the migration from Pro to Essentials.
Now that we are thinking about supporting other use cases, we should make it more generic:
- Rename `EssentialsPlan` to `MigrationPlan`
- Use current plan name instead of hardcoded `Publish`
- Use suggested plan name instead of `Essentials`
- Added a loading state to be able to use data

Everything should look the same from a user point of view.
https://www.loom.com/share/d4563a1ff8c4458c94f4bed1df05ac9f


jacobchadwell (approved):
🚀 looks good to me!

PR #424, opened 2022-09-27
[UPDATE] Pass current plan not selected for tracking

daisymarie128:
👀 Would I be able to get a review on this one? As I made the ticket and just want to check this is the right value which should be passed back 🙂

[GROW-715](https://buffer.atlassian.net/browse/GROW-715)

This PR fixes the `currentPlan` value we send to the tracking event `Plan Selector Viewed`

Also in this PR:
Adds missing props to component for Revenucat customers - The Exeperminet which uses this component is not turned on yet so we haven't seen any problems with this yet. However, when we turn it on we would have been missing some of these values.

Previously we were sending the user's current selected plan in the Modal. Which does not match what our tracking event describes this value should be: 
[Tracking definition for currentPlan](https://github.com/bufferapp/tracking-plan/blob/master/tracking-plans/buffer-tracking-plan/events/product/shared/plan-selector-viewed.yaml#L11)

**The image shows my users on Prod which is on a Free Plan but the data collected indicates I am on a teams plan which is not true**
Screen Shot 2022-09-27 at 9 22 29 am


jacobchadwell (approved):
This looks great to me! 🎉

PR #427, opened 2022-10-10
[GROW-706] Check shouldShowPaywall on each render to enforce entitlem…

jacobchadwell:
…ents

https://buffer.atlassian.net/browse/GROW-706
See context and great video walkthrough by Martin in the ticket above ^

This change performs the check for `shouldShowPaywall` on each render. Before, the function was not triggering on organization switches because the useEffect it was inside had `user.loading` as its dependency array. Org switching writes the updated account query results to the cache, but never actually triggers another loading state, so this function was never called upon org switch.

(I feel like removing functions from `useEffects` will be more [common](https://beta.reactjs.org/learn/you-might-not-need-an-effect)... especially if we ever upgrade to React 18)


cheahenry (approved):
LGTM!
It shouldn't add much load having it at every render because it's just [checking the current url and state](https://github.com/bufferapp/app-shell/blob/c40ded011cee0083ffb76ac90afecf0694c08111/packages/app-shell/src/components/Modal/utils.js#L64-L70).

PR #447, opened 2022-11-21
[GROW-750] Add legacy migration CTAs

jacobchadwell:
Add two split-controlled migration CTAs to the NavBar (one in the menu, one on the nav bar itself)

image


packages/app-shell/src/components/NavBar/utils/utils.js
```diff
@@ -0,0 +1,15 @@
+// TODO: move additional functions here from NavBar index
```
jacobchadwell, line 1: Moved some utility functions here to avoid a circular dependency. We can move more of the URL utilities here later if we want

PR #448, opened 2022-11-21
Revert "[GROW-750] Add legacy migration CTAs (#447)"

jacobchadwell:
This reverts commit c3f5578ac965ec1faaf7ec0e28bdbad59d40f31a.


jacobchadwell:
I was seeing a strange flashing of the CTA even for users in the "Off" treatment group. Need to revisit the loading/rendering order of the CTA

jacobchadwell:
Just to leave an update - I wasn't able to reproduce this issue, and I've tested again with the same code in prod with no issue. I think I just needed a hard refresh on my browser since I was playing around with lots of splits 😄

PR #461, opened 2023-01-10
[EPD-653] cleanup and changes to the help menu

cmunozgar:
Context: https://buffer.atlassian.net/browse/EPD-653

CleanShot 2023-01-10 at 22 53 40


mickmahady (approved):

packages/app-shell/src/components/NavBar/utils/dropdown-items.test.js
```diff
@@ -0,0 +1,21 @@
+import { getHelpDropdownItems, getAppsDropdownItems } from './dropdown-items';
+
+describe('getHelpDropdownItems', () => {
+  it('returns the correct number of items in the list', () => {
```
gisete, line 4: hey @cmunozgar (ceiling 😸  here) and just a minor comment but what do you think if we were a little more specific in the description so we can differentiate these 2 tests? 🙂 
"_...items in the apps dropdown_" and "..._items in the help dropdown_"?

cmunozgar: Thank you for spotting that @gisete 🌮

PR #525, opened 2023-02-16
[NEW] Code Climate Config

daisymarie128:
This PR adds a `.codeclimate.yml` config file to our repo.

Main purpose of this is to exclude test files and other unnecessary files which shouldn't be included in reports.

This config also turns on the eslint plugin (testing to see if this is useful)


jacobchadwell:
Looks great! Thanks for adding this @daisymarie128 !!

jacobchadwell (approved):

PR #536, opened 2023-02-21
Task/grow 1025 restructure tracking

daisymarie128:
This PR starts the work for GROW-1025

It creates a new folder structure where we can create and define tracking events.

This PR moves 2 tracking events for the PlanSelector:
1. trackPlanSelectorViewed
2. trackPlanClicked

The Folder structure is as follows

```md
.
├── ...
├── src               
│ ├── components            # Is on the same level as components, and the common folder
│ └── tracking            # This is where we will now create and test all tracking events we need to use
│   ├── PlanSelector             # Is on the same level as components, and the common folder
│    ├── PlanSelectorTrackingTypes.ts             # Definition of types
│    ├── trackPlanSelectorClicked.test.ts             # All tracking files follow naming trackNameOfEvent etc 
│   └── trackPlanSelectorClicked.ts            # All tracking functions have their own file
│  └── utils.ts              # Common funcs used between all tracking events  
└── ...
```

## Folder Structure
Screen Shot 2023-02-21 at 10 40 42 am

**Tracking event still being called after refactor**
Screen Shot 2023-02-21 at 10 22 44 am
Screen Shot 2023-02-21 at 10 23 00 am

**All tests new and old passing**
Screen Shot 2023-02-21 at 10 24 17 am


packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.jsx
```diff
@@ -1,9 +1,20 @@
+// Note: All these fixes need to be removed soon. They are here to reduce the noisy errors being ignored when we run our CIs
```
daisymarie128, line 1: I've added disables to this file type to help remove some of the errors showing in our CI checks.

These errors are relevant though and **SHOULD** be fixed. They are only being ignored because we have recently added TS to the repo and fixing all these type definitions would take a considerable amount of time and we still want to be able to move fast with our pace.

By adding the FIXME comment this tech debt can now be tracked in codeclimate and we can still be alerted to these comments reduce our code quality.

PROPOSAL:
We slowly create tickets to commit to fixing these comments 1 at a time throught our cycles.
We do not add more ignores to files with these definitions. When we add new code to a file we should follow the linting rules.

cheahenry: That sounds like a good plan, thanks Daisy!

jacobchadwell: @daisymarie128 I love this idea of incrementally adopting TS and keep shipping speed up 🚀

jacobchadwell:
Also love this new folder structure! 🤩

PR #550, opened 2023-03-01
[GROW-1036] Fix PlanSelectorContainer type errors

jacobchadwell:
Fix type errors on the PlanSelectorContainer


packages/app-shell/src/components/Modal/modals/PlanSelector/components/PlanSelectorContainer.tsx
```diff
@@ -368,15 +363,11 @@ export const PlanSelectorContainer = ({
           {shouldShowDowngradeToFree &&
             (shouldIncludeAgencyPlan || showAgencyPlan) && (
               <FreePlanSection
-                ctaAction={() => {
+                ctaAction={(): void => {
                   updateSelectedPlan(
                     `${freePlan.planId}_${monthlyBilling ? 'month' : 'year'}`
                   );
-                  updatePlan({
```
jacobchadwell, line 375: Similar to the <Error> component above, we never actually used these arguments passed into `updatePlan`. The arguments instead are defined inside the hook.

PR #552, opened 2023-03-02
[UPDATE] Add new hasNeverBeenPayingCustomer to query

daisymarie128:
# Description

This PR adds the newly created `hasNeverBeenPayingCustomer` property to the Billing query

This is ready to be used to handle whether or not to take a customer to the Stripe Checkout and turn on automatic tax collection

**Value coming back correctly for different user states**
Screen Shot 2023-03-02 at 4 50 39 pm

Screen Shot 2023-03-02 at 4 48 04 pm


hamstu:
@josem Is this still an issue for you?

daisymarie128:
@josem Sorry for the late reply, I think this may have needed a pull to get the latest from buffer-mono. Let me know if this is still causing issues 🙂

PR #566, opened 2023-03-14
[GROW-1072] Fix remaining TS errors blocking migration

jacobchadwell:


packages/app-shell/src/common/hooks/useModal.ts
```diff
@@ -79,9 +79,13 @@ function useModal({ refetchAccountData }: UseModalProps): UseModalReturn {
   }
 
   useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
jacobchadwell, line 83: I'm ignoring the errors on the event listeners for now - we know they work, and we will likely get rid of these anyway if we directly consume the app-shell in buffer-mono

packages/app-shell/src/components/Modal/modals/PlanSelector/hooks/types.ts
```diff
@@ -1,12 +1,13 @@
 import { ErrorState } from 'common/types/graphQL/Error';
 import { ChangePlanOption } from 'common/types';
+import { ModalData } from 'common/events/types';
 
 export type UseCheckoutSession = {
-  organizationId: string;
+  organizationId?: string;
```
jacobchadwell, line 6: We handle the case where organizationId is undefined 👍

PR #582, opened 2023-03-27
[Migration Prep] Eject react scripts

jacobchadwell:
Test PR to address issue found during migration (known issue with create-react-app - https://github.com/facebook/create-react-app/issues/11865).

Because we will not need to make updates to the test app (eventually deprecating entirely in the monorepo), I'm going to eject the configs so that I can make the small change to webpack needed to successfully migrate to buffer-mono.


jacobchadwell:
Not working - may try and just get Hermes working with app-shell instead.

PR #587, opened 2023-03-30
Fix NavBar product links based on environment

cmunozgar:
This PR uses `getUrlEnvModifier` to set the product links according to the environment. This works on `prodution`, `staging` and `local` environments.


mickmahady (approved):
This is amazing! 🌮 🌮

